PR #6, opened 2018-03-07
add paperwallet glue code

vincenthz:


js/PaperWallet.js
```diff
@@ -1,4 +1,40 @@
...
+export const scramble = (module, iv, password, input) => {
+    if (iv.length != 4) {
+        throw new Error('IV must be 4 bytes');
+    }
+    const bufiv = newArray(module, iv);
+    const bufinput = newArray(module, input);
+    const bufpassword = newArray(module, password);
+    const bufoutput = newArray0(module, input.length + 4);
+    module.paper_scramble(bufiv, bufpassword, password.length, bufinput, input.length, bufoutput);
+    let result = copy_array(module, bufoutput, input.length + 4);
```
DominikGuzei: `copy_array` will not work here, since it's now called `copyArray`

js/PaperWallet.js
```diff
@@ -1,4 +1,40 @@
+import { newArray, newArray0, copyArray } from './utils/arrays';
```
DominikGuzei: We don't have `newArray0` in the utils … i refactored it to a single `newArray` method that can be called with `true` as last param to indicate that it should nullify.

vincenthz: newArray0 take an implicit size, not assume it from the input parameter. it doesn't work in this case. (one might say that newArray was miscalled in the first place and should be a copyArray like name)

js/PaperWallet.js
vincenthz: good catch. looking forward to be able to test everything smoothly

js/utils/arrays.js
```diff
@@ -9,6 +9,16 @@ export const newArray = (module, b, isZero=false) => {
   return ptr
 }
 
+export const newArray0 = (module, len) => {
```
DominikGuzei: Haha, ok … you really want this duplicated code i guess ;) please look at the `newArray` function, then you will see what i meant … this is already implemented there

js/PaperWallet.js
DominikGuzei: Ah ok… sorry then i misunderstood how it works … although the implementation looked 99% the same apart from this single param

js/utils/arrays.js
vincenthz: how do you expect `b.length` work when b doesn't exist ?

PR #8, opened 2018-03-08
[DDW-130] integrate with mnemonics example

DominikGuzei:
This PR integrates the mnenomics app example with the proper js api that will be used by external consumers like Daedalus. It also removes any unnecessary files (like the haskell server)


NicolasDP:
The commit message is a bit understating all the changes made. I don't understand why the `js-example` has been touched here. This will conflict with other WIP for no good reasons.

DominikGuzei:
@NicolasDP sorry for that, i wasn't aware of other WIP and wanted to cleanup the codebase. `js-example` was simple renamed to `example` and slightly restructured (nothing serious). All the purescript parts etc still work the same way (except that they now properly integrate with the generated JS api that is exposed to the outside world)

DominikGuzei:
@NicolasDP i renamed it back to `js-example` 😉

DominikGuzei:
@NicolasDP and @vincenthz this is ready to be reviewed again.

README.md
```diff
@@ -22,3 +22,20 @@ to be installed prior to running this.
...
+
+Running the Example
+-------------------
+There is a simple example application in `js-example` that can be run to test some of the features.
+
+### installation
+
+within `js-example/` folder
+
+2. `npm i && npm i -g bower`
```
darko-mijic: It would be better to add bower as an NPM dev dependency instead of installing it globally. Then it can be used from `node_modules` or from the script in `package.json`the same way as if it was installed globally. @DominikGuzei

darko-mijic:
```
"scripts": {
...
    "bower-install": "bower install"
  }
```

`npm run bower-install`

DominikGuzei: Yeah that's right … will update

PR #13, opened 2018-03-09
Fix paper wallets api

DominikGuzei:
This PR adds the folowing two methods to the `CardanoCrypto.PaperWallet` api:

`scrambleStrings: (iv:Uint8Array, password:string, mnemonics:string) -> 15 words`
`unscrambleStrings: (password:string, shielded_input:string) -> 12 words`

and implements all test vectors [outlined here](https://github.com/input-output-hk/cardano-specs/blob/master/proposals/0001-PaperWallet.md)


js-example/index.html
```diff
@@ -23,4 +23,12 @@
...
   <script src="../dist/index.js"></script>
   <script src="output/index.js"></script>
+  <script>
+    setTimeout(function() {
+      const iv = new Uint8Array(4);
+      window.crypto.getRandomValues(iv);
+      const result = CardanoCrypto.PaperWallet.scramble(iv, '', "legal winner thank year wave sausage worth useful legal winner thank yellow");
+      console.log(result);
+    }, 1000)
+  </script>
```
NicolasDP: The expected result here is:

`abandon abandon abandon win nest chef want salt join shove minor december miss oak name`

can this bit be moved to a test suite in `js` directory ? Might be worth adding unit tests with all the test vectors.

js/Bip39.js
```diff
@@ -0,0 +1,18 @@
...
+import Blake2b from './Blake2b';
+import { apply } from './utils/functions';
+
+export const mnemonicToSeed = (module, mnemonic) => (
+  Blake2b.blake2b_256(bip39.mnemonicToEntropy(mnemonic))
+);
+
+export const mnemonicToEntropy = (mnemonic) => (
+  Buffer.from(bip39.mnemonicToEntropy(mnemonic), 'hex')
+);
```
NicolasDP: It might be worth checking how all of this can work with what is already with daedalus:

https://github.com/input-output-hk/daedalus/blob/b85ad0e212f22236d6404dd3b7c7a9fe197f410c/lib/decrypt.js

Since Daedalus already has already [a function to retrieve the entropy from a mnemonic](https://github.com/input-output-hk/daedalus/blob/b85ad0e212f22236d6404dd3b7c7a9fe197f410c/lib/decrypt.js#L33) might be worth keeping using the entropy here. and not adding this extra Bip39 handling.

js-example/index.html
DominikGuzei: Yes, moving this into a js test suite is another goal 😉

DominikGuzei:
@NicolasDP @vincenthz this PR is now done and ready for re-review by you guys! I implemented test cases for all test vectors which can be run with `npm run test`. Travis integration is still missing but can be added separately in a later PR 😉

PR #22, opened 2018-03-26
[DDW-174] also run js test suite on travis

DominikGuzei:
The target of this PR is to make node / JS test suite run on travis additionally to the current rust tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ matrix:
     - rust: stable
     - rust: beta
     - rust: nightly
+    - rust: nightly
+      env: NODE_JS_GLUE
       before_script:
         - rustup target add wasm32-unknown-unknown --toolchain nightly
         - nvm install 8.2.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,15 +4,18 @@ rust:
   - stable
   - beta
   - nightly
-#matrix:
-#  allow_failures:
-#    - rust: nightly
-before_install:
-  - nvm install 8.2.1
-before_script:
-  - npm install
 script:
   - cargo build --verbose --all
   - cargo test --verbose --all
-  - ./build
-  - npm run test
+matrix:
+#  allow_failures:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,9 +4,18 @@ rust:
...
   - nightly
-#matrix:
-#  allow_failures:
-#    - rust: nightly
 script:
   - cargo build --verbose --all
   - cargo test --verbose --all
+matrix:
+#  allow_failures:
+#    - rust: nightly
```
NicolasDP: can you remove these comments as well? They are no longer needed anyway. We need to build on nightly.

DominikGuzei: Yeah sure 😉 should i remove the other rust build targets (stable, beta)?

DominikGuzei:
@NicolasDP there are some new javascript tests which are failing (also on my local machine) … is that intentional?

NicolasDP:
yes, we were creating addresses with a single key distribution, which is not the current scheme. This has been fixed but not the test vector at the JS level. I will push something soon about it

DominikGuzei:
@NicolasDP ok, this PR is done though - the tests are running correctly for the rust `nightly` build. Please review and merge!

.travis.yml
```diff
@@ -1,12 +1,17 @@
...
-#  allow_failures:
-#    - rust: nightly
 script:
   - cargo build --verbose --all
   - cargo test --verbose --all
+matrix:
+  include:
+    - rust: stable
+    - rust: beta
+    - rust: nightly
```
NicolasDP: duplicating this line and adding a env variable `TEST_NODEJS_BINDINGS` would be enough and would show in the travis-ci board

DominikGuzei:
@NicolasDP can you make the necessary changes - im on holidays until 4th of April!

NicolasDP:
I had to do a rebase of the code to master in order to fix the travis script.

PR #45, opened 2018-04-17
Expose Rust load module method

mirkoAlic:
The goal of this PR is to expose the `loadRustModule` method in order to be able to call it without configuring web assembly loader within the container app webpack.

We need to expose `loadRustModule` in order to be able to wait until crypto lib is loaded before doing any other operation.


NicolasDP:
thanks, looking good

PR #49, opened 2018-04-19
Ready to use tx

### LINKED ISSUE #41: Ready made transaction to send to network
From UTXO & Sending Targets, create a valid transaction signed and with correct fees

NicolasDP:
closes #41 

however, this is still block by #28 where we need to be able to generate the address for the change.


NicolasDP:
regarding #28 I just pushed something that may not be correct and will need reviewed.

PR #55, opened 2018-04-23
Add shell.nix file for building at least the rust part.

nc6:


shell.nix
```diff
@@ -0,0 +1,16 @@
+let
+  moz_overlay = import (builtins.fetchTarball https://github.com/mozilla/nixpkgs-mozilla/archive/master.tar.gz);
+  nixpkgs = import <nixpkgs> { overlays = [ moz_overlay ]; };
+in
+  with nixpkgs;
+  stdenv.mkDerivation {
+    name = "moz_overlay_shell";
+    buildInputs = [
+      (nixpkgs.latest.rustChannels.nightly.rust.override { targets = ["wasm32-unknown-unknown"]; })
```
NicolasDP: it is not necessary to build with nightly for wasm32. the project can compile in stable and unstable too. It depends of what rust users wants to do : hence it is better to use `rustup` or to specify the build options with `cargo`.

NicolasDP: thanks, this is going to be addressed in #58

vincenthz:
I think we don't need this anymore with #58 merged

PR #58, opened 2018-04-24
remove all JS relating files

NicolasDP:
they are all moved to [js-cardano-wasm](https://github.com/input-output-hk/js-cardano-wasm)


vincenthz:
@darko-mijic @DominikGuzei something to be aware of, we've moved the js to its own repo, so we can simplify the integration, testing of each parts

Issue #90, opened 2018-05-21
block genesis finish parsing / printing


vincenthz:
Finish parsing of genesis blocks (body), and oretty printing of block

plredmond:
are you open to using a library for the pretty printing, such as https://docs.rs/pretty/0.4.0/pretty/ ?

plredmond:
Ok, so there's a generic prettyprinter now. Next steps:

- [ ] parse genesis blocks
- [ ] implement `Pretty` for genesis blocks

PR #101, opened 2018-05-24
Generic pretty-printer and block transformation implementations

plredmond:
* First commit adds a generic pretty-printer for nested key-value pairs with optional key colors and configurable indentation. There's also a `fmt::Display` implementation which forgets the colors.
* Second commit is a ton of `impl Pretty` to replace about 40 lines of `println!`. It could be implemented more concisely if I eliminated `trait Pretty` and wrote a conversion for each concrete type you'd like to print. Which do you prefer, traits or concrete & concise?

Part of #90

### DIFF
--- a/blockchain/Cargo.toml
+++ b/blockchain/Cargo.toml
@@ -13,7 +13,6 @@ rcw = { path = "../rcw" }
 wallet-crypto = { path = "../wallet-crypto" }
 serde = "1.0"
 serde_derive = "1.0"
-ansi_term = "0.9"
 
 [features]
 with-bench = []
--- a/blockchain/Cargo.toml
+++ b/blockchain/Cargo.toml
@@ -13,6 +13,7 @@ rcw = { path = "../rcw" }
 wallet-crypto = { path = "../wallet-crypto" }
 serde = "1.0"
 serde_derive = "1.0"
+ansi_term = "0.9"
 
 [features]
 with-bench = []
--- a/blockchain/src/lib.rs
+++ b/blockchain/src/lib.rs
@@ -16,5 +16,8 @@ pub mod genesis; /* genesis block related value */
 pub mod normal; /* normal block related value */
 mod block;
 
+pub mod pretty;
+extern crate ansi_term;
+
 pub use types::*;
 pub use block::*;
--- a/blockchain/src/lib.rs
+++ b/blockchain/src/lib.rs
@@ -16,8 +16,5 @@ pub mod genesis; /* genesis block related value */
 pub mod normal; /* normal block related value */
 mod block;
 
-pub mod pretty;
-extern crate ansi_term;
-
 pub use types::*;
 pub use block::*;
--- a/blockchain/src/pretty.rs
+++ b/blockchain/src/pretty.rs
@@ -4,6 +4,12 @@ use std::string::String;
 
 use ansi_term::Colour;
 
+use block;
+use genesis;
+use normal;
+use types;
+use wallet_crypto;
+
 static DISPLAY_INDENT_SIZE: usize = 4; // spaces
 static DISPLAY_INDENT_LEVEL: usize = 0; // beginning starts at zero
 static DISPLAY_USE_COLOR: bool = false; // no color for display implementations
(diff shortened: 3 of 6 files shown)


plredmond:
# old image
# new
image

blockchain/src/pretty.rs
```diff
@@ -0,0 +1,405 @@
...
+        let mut input = Vec::new();
+        input.push(("character".to_string(), Pairs(Some(Colour::Blue), nested)));
+        input.push(("crook".to_string(), Single(Box::new("yes"))));
+        assert_eq!(
+            format!("{}", Pairs(Some(Colour::Red), input)),
+            "\
+- character:
+    - name: zaphod
+    - age : 42
+- crook    : yes
```
NicolasDP: I could vote for him

blockchain/src/lib.rs
```diff
@@ -16,5 +16,8 @@ pub mod genesis; /* genesis block related value */
 pub mod normal; /* normal block related value */
 mod block;
 
+pub mod pretty;
+extern crate ansi_term;
```
NicolasDP: keep the `extern crrate` lines before the `mod` of the library.

blockchain/src/pretty.rs
```diff
@@ -0,0 +1,405 @@
...
+
+static DISPLAY_INDENT_SIZE: usize = 4; // spaces
+static DISPLAY_INDENT_LEVEL: usize = 0; // beginning starts at zero
+static DISPLAY_USE_COLOR: bool = false; // no color for display implementations
+
+type Assoc = (String, Val);
+
+// core pretty-printer type which facilitates homogenous ordered key-value pairs
+pub enum Val {
+    Single(Box<fmt::Display>),
```
NicolasDP: one could use [`fmt::Argument`](https://doc.rust-lang.org/std/fmt/struct.Arguments.html) here and remove the `Box` This way we are not link to `fmt::Display` which has a specific meaning in rust, different from the pretty printing we want to use.

A good example of usage is the crate [log](https://github.com/rust-lang-nursery/log) which uses the `fmt::Argument` structure instead of the `fmt::Display` trait so the user can use any formatting trait (Debug, hexa...).

blockchain/src/pretty.rs
```diff
@@ -0,0 +1,405 @@
...
+) -> fmt::Result {
+    match p {
+        // format pretty-val as a terminal
+        Val::Single(display) => write!(f, "{}", display),
+        // format pretty-val as a set of  key-vals
+        Val::Pairs(color, pairs) => {
+            let key_width = longest_key_length(pairs);
+            pairs
+                .iter()
+                .fold(Ok(()), |prev_result, (key, val)| match prev_result {
```
NicolasDP: avoid the match on the `Result` here:

```rust
prev_result.and_then(|()| {
  let col = ...;
  ...
})
```

blockchain/src/pretty.rs
```diff
@@ -0,0 +1,405 @@
...
+
+type Assoc = (String, Val);
+
+// core pretty-printer type which facilitates homogenous ordered key-value pairs
+pub enum Val {
+    Single(Box<fmt::Display>),
+    Pairs(Option<Colour>, Vec<Assoc>),
+}
+
+// XXX: this class is related to "into" somehow
```
aspiwack: It does, formally, but I don't think this is particularly useful a remark. `Into` is used as a generic conversion mechanism in macros. But you probably don't want to convert implicitly to a `Pretty` anyway. Plus it is recommended to not implement `Into` directly, but to implement `From` instead. Which is impractical for this use-case.

blockchain/src/pretty.rs
```diff
@@ -0,0 +1,405 @@
...
+}
+
+// format any Val to a string with color
+pub fn format_val(p: &Val, indent_size: usize) -> String {
+    // internal helper to get a formatter for calling `fmt_pretty`
+    struct Fmt<F>(F)
+    where
+        F: Fn(&mut fmt::Formatter) -> fmt::Result;
+
+    impl<F> fmt::Display for Fmt<F>
```
aspiwack: The use of this local `Display` instance is suspicious. Unfortunately I don't have any suggestion to improve on this.

blockchain/src/pretty.rs
```diff
@@ -0,0 +1,405 @@
...
+        // format pretty-val as a terminal
+        Val::Single(display) => write!(f, "{}", display),
+        // format pretty-val as a set of  key-vals
+        Val::Pairs(color, pairs) => {
+            let key_width = longest_key_length(pairs);
+            pairs
+                .iter()
+                .fold(Ok(()), |prev_result, (key, val)| match prev_result {
+                    // return an error
+                    err @ Err(_) => err,
```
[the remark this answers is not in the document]
plredmond: no, just inexperience with rust.. how do you exit the iter on error? woudl using `and_then` as @NicolasDP suggested do it like in the monad world?

blockchain/src/pretty.rs
plredmond: thanks for explaining!

blockchain/src/pretty.rs
plredmond: it was necessary to get an instance of `fmt::Formatter` but @NicolasDP suggested an alternative structuring.. thanks!

blockchain/src/pretty.rs
plredmond: if i switch to using the `fmt.Arguments` struct, then would i be using the `format_args!` macro every time i construct a `Single`?

blockchain/src/pretty.rs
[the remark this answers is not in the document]
aspiwack: If you encounter an error, I assume that you actually want to return from the function with an error, which you can achieve by using `return Err` instead of `err`.

That's what the `?` notation does for you: `a?` is equivalent to
```rust
match a {
  err @ Err(_) => return err, -- this returns immediately from the function
  Ok(a') => a'
}
```
It's quite convenient, if you don't intend to manage the error in the same function.

blockchain/src/pretty.rs
NicolasDP: yes that would be the idea.

blockchain/Cargo.toml
```diff
@@ -13,6 +13,7 @@ rcw = { path = "../rcw" }
 wallet-crypto = { path = "../wallet-crypto" }
 serde = "1.0"
 serde_derive = "1.0"
+ansi_term = "0.9"
```
NicolasDP: just noticed, blockchain does not need to depend on `ansi_term`

blockchain/src/pretty.rs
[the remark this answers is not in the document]
plredmond: I think `=> err` returning from the closure on line 97.

Would `=> return err,` be any different? What are you saying itnwould return from?

plredmond:
Thanks for the comments @nicolasdp! I'll update the pr, but I'm confused about one thing you said:
> Ideally the AST should have more constructor: Hash, Address, Coin, etc...
I'd like to clarify the requirement because @vincenthz asked for the pretty-printer to be generic and indicated the structures could be handled as key-value pairs. Let me know how to proceed!

blockchain/src/pretty.rs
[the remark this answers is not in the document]
aspiwack: I thought so, but I was wrong. So, yeah, using `?` here would not be different from using `and_then`.

vincenthz:
yes @NicolasDP is correct, but this is additional values we can add later.

the way I see the pipeline is that all types we interested about can generate an AST like:

```
type AST = [(Key, Value)]
data Value = String data | Tree AST 
```

so as a pipeline: `a => AST => Pretty Printer`

then later we can add specialized values which has special printer and associated configurable colors

blockchain/src/pretty.rs
plredmond: I'm not entirely sure how the log package you linked me to works, but [i'm encountering an error about a dropped borrowed value](http://play.rust-lang.org/?gist=d3e687087ab418cdda66e4ed14cdf78a&version=undefined&mode=undefined) whenever I attempt to save the result of `format_args!`. If I'm missing something, please let me know. However, in light of that I'll probably use `String` for the plain-value case.

plredmond:
* @NicolasDP I was unable to implement the version taking `fmt::Arguments` ([comment](https://github.com/input-output-hk/rust-cardano/pull/101#discussion_r190781995)) so I switched from `Box<..>` to `String` as @vincenthz suggested.

    * ~you don't need to use Box<fmt::Display>; use directly the fmt::Argument object, Display has a specific meaning, here we don't want to enforce a meaning;~
    * ~use `std::io::Write` instead of `fmt:Formatter`~
    * [x] use `String` instead of `Box<fmt::Display>`

* @NicolasDP Making the tree rich with blockchain types is currently underspecified and @vincenthz indicated a simple tree is desirable for now.

    * ~Ideally the AST should have more constructor: Hash, Address, Coin, etc...~
    * ~The current pretty printing is terrible, we would not mind a complete new makeover.~

* These changes were no problem

    * [x] move pretty-printer from `blockchain` to `wallet-cli`
    * [x] remove color from AST

* New output

image

PR #109, opened 2018-05-28
More pretty printer features

plredmond:
implementing aux requests noted in [a comment in #90](https://github.com/input-output-hk/rust-cardano/issues/90#issuecomment-392023896); also see commit messages

  - [x] add color to `previous_hash` (field of `BlockHeader`)
  - [x] add color to `block_signature` (field of `Consensus`)
  - [x] extend the AST to allow `Val::List<Vec<Value>>`, use it for lists like transactions and the long list in genesisblocks
  - [x] implement `Pretty` for genesis blocks (use same approach as `blockchain::normal::Block` implementations)

## main block

image

## genesis block

image

### DIFF
--- a/blockchain/src/genesis.rs
+++ b/blockchain/src/genesis.rs
@@ -123,7 +123,7 @@ impl cbor::CborValue for Block {
 
 #[derive(Debug, Clone)]
 pub struct Consensus {
-    pub epoch: u32,
+    pub epoch: types::EpochId,
     pub chain_difficulty: ChainDifficulty,
 }
 impl cbor::CborValue for Consensus {
--- a/blockchain/src/lib.rs
+++ b/blockchain/src/lib.rs
@@ -11,7 +11,7 @@ extern crate test;
 extern crate serde_derive;
 extern crate serde;
 
-mod types;
+pub mod types;
 pub mod genesis; /* genesis block related value */
 pub mod normal; /* normal block related value */
 mod block;
--- a/wallet-cli/README.md
+++ b/wallet-cli/README.md
@@ -27,7 +27,7 @@ jYTLseJK1m2RmsXD44EZVe2EqGtBnQyFnPwjS2Q36c3iLwWCPGmAVWKR9ufE
 jYTLseJK1m2eYuy16w7tjPNqtVdAaQ5bnb4Cyya2YH9AqJoNPQ1VLgE4MaXL
 ```
 
-Make a network configuration (from a template).
+Make a blockchain network configuration (from a template).
 
 ```sh-session
 $ ariadne blockchain new foo --template testnet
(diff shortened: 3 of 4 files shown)


vincenthz:
This look fine apart from the colour handling. the idea is to have high level values in the AST like `Number` and `Hash`,  that will be rendered at the print site with color, hex escape, etc.. depending on their ast type.

plredmond:
- handle colors with additional `Val` constructors
    - [ ] construct cyan `Val::Hash` for `previous_hash`

@vincenthz i have two questions:
1. what are some other components which require their own constructor? you mentioned `Number`, is that just any `i32`/`i64`/`u32`/`u64` value?
2. block-signature is a rich value with many subcomponents; how would you like it to be colored?

NicolasDP:
I think I can help with that:

1. I can think of these: Addresses; hashes, the Epoch/Slotid; Signatures, public keys
2. There might be some special objects relevant from the point `1.` (signature, hashes, pk), that might be enough for now. We can enhance it later.

plredmond:
Checking out the cardano code reveals what I was misunderstanding:

1. [Pos/Core/Common/Types.hs#L81-L85](https://github.com/input-output-hk/cardano-sl/blob/b72436850ce92f429d7942b5b5e7cd9cb510a63e/core/src/Pos/Core/Common/Types.hs#L81-L85)
1. [Pos/Crypto/Hashing.hs#L86-L90](https://github.com/input-output-hk/cardano-sl/blob/b72436850ce92f429d7942b5b5e7cd9cb510a63e/crypto/Pos/Crypto/Hashing.hs#L86-L90)
1. [Pos/Explorer/Web/ClientTypes.hs#L91-L105](https://github.com/input-output-hk/cardano-sl/blob/b72436850ce92f429d7942b5b5e7cd9cb510a63e/explorer/src/Pos/Explorer/Web/ClientTypes.hs#L91-L105)
    ```haskell
	-- See this page for more explanation - https://cardanodocs.com/cardano/addresses/
	-- We have the general type @AbstractHash@ for all hashes we use. It's being parametrized
	-- by two types - AbstractHash algo a - the hashing algorithm and the phantom type for
	-- extra safety (can be a @Tx@, an @Address@ and so on, ...).
	--
	-- The following types explain the situation better:
	
	type AddressHash   = AbstractHash Blake2b_224
	type Hash          = AbstractHash Blake2b_256
	
	type TxId          = Hash Tx               = AbstractHash Blake2b_256 Tx
	type StakeholderId = AddressHash PublicKey = AbstractHash Blake2b_224 PublicKey
    ```

@NicolasDP From this I suppose it would make sense for us to have
1. `Val::Hash` ..?
1. `Val::AddressHash` ..?
1. `Val::TxId` -- used for transaction ids
1. `Val::StakeholderId` -- used for addresses/publickeys

NicolasDP:
You don't need AddressHash, it is hidden in the `ExtendedAddr` and we prefer to display the extended address as base58 all serialised in the appropriate format.

While TxId would be useful, it is not in the data structure to be displayed. But it is still a Hash, so no need to duplicate it in its special sum_type constructor.

PR #112, opened 2018-05-30
Parse more genesis block data

### LINKED ISSUE #90: block genesis finish parsing / printing

plredmond:
* adddce6 parse stakeholder-ids out of `genesis::Block` cbor
    * is there more data to parse for `genesis::Block`?
* 8ee76a0 add pretty-printer constructor `Val::Stakeholder` and formatter case which colors it
    * gb
* probably resolves #90

### DIFF
--- a/blockchain/src/genesis.rs
+++ b/blockchain/src/genesis.rs
@@ -1,3 +1,4 @@
+use std::collections::LinkedList;
 use wallet_crypto::{address, cbor, hash::{Blake2b256}};
 use wallet_crypto::cbor::{ExtendedResult};
 use wallet_crypto::config::{ProtocolMagic};
@@ -20,7 +21,7 @@ impl cbor::CborValue for BodyProof {
 
 #[derive(Debug, Clone)]
 pub struct Body {
-    pub slot_leaders: Vec<address::StakeholderId>,
+    pub slot_leaders: LinkedList<address::StakeholderId>,
 }
 /*
 impl fmt::Display for Body {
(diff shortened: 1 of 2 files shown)


blockchain/src/genesis.rs
```diff
@@ -36,9 +34,19 @@ impl cbor::CborValue for Body {
...
     fn decode(value: cbor::Value) -> cbor::Result<Self> {
-        value.iarray().and_then(|array| {
-            Ok(Body { slot_leaders: array })
-        }).embed("While decoding genesis::Body")
+        Ok(Body {
+            slot_leaders: value
+                .iarray()
+                .and_then(|array| {
+                    array.iter().fold(Ok(vec![]), |acc, item| {
+                        acc.and_then(|mut vec| {
```
NicolasDP: I think this is a bit far fetch there. The fold on an error here is not necessary and is making the code a bit complex to read.

There is an instance of `impl<T> CborValue for LinkedList<T> where T: CborValue`. I agree it is not great to use LinkedList (poor perf and co), but the easy thing to do is:

#### keep the LinkedList in the code

```rust
pub struct Body {
  pub slot_leaders: LinkedList<address::StakeholderId>
}
```

#### leverage the instance of linkedlist

```rust
fn decode(value: cbor::Value) -> cbor::Result<Self> {
    let slot_leaders = cbor::CborValue::decode(value).embed("...")?;
    Ok(Body { slot_leaders })
}
```

This will do all the necessary work for you.

PR #126, opened 2018-06-02
Additional parsing for normal blocks

plredmond:
This implements `SccPayload` decoding & pretty printing for normal blocks. It might make sense for some of the new types to be moved to `blockchain::types` if genesis blocks need them.

Encoding isn't implemented.

These are the blocks I'm testing the decoding & pretty-printing against:
```shell
echo === four constructors of SccPayload
cargo run -- blockchain cat mainnet 35481e8ed589f800ebcbc596aaf06d0e5c55b8f6af9d8b134eef03bfdc198199
cargo run -- blockchain cat mainnet 65ce050df93f883f521ec91756393286d509886177d187c8f099e4fc3f0fb7cc
cargo run -- blockchain cat mainnet b05742df56b52c2fbdb9daba0a8d93bd9d1a8f8ff0185922f544b7234460bba6
cargo run -- blockchain cat mainnet 3d1f0f015452cb0266fdbbcc8f09a5dd144fd188f02049977314bc205ab5166f
echo == non empty vss in SccPayload
cargo run -- blockchain cat testnet f134770887ff970642f215bee0d5465e1fa8fa1f8c052c38709c42c69e1dcad5
cargo run -- blockchain cat testnet 629f3fc87e48af8739c94333eba2bbc8832e7f2ffb2688cd58cbfefade467b64
cargo run -- blockchain cat testnet 3e726761ee51a9cc0fafbe8108c9e460afb40f0ba9fd69005f8da5ea511ed073
```


plredmond:
Ugh. `rustfmt` made this code a bit unreadable. Sorry.

wallet-crypto/src/cbor/spec.rs
```diff
@@ -276,6 +276,15 @@ pub enum ObjectKey {
...
 }
+impl ObjectKey {
+    pub fn value(self) -> Value {
+        match self {
+            ObjectKey::Integer(v) => Value::U64(v),
+            ObjectKey::Bytes(v) => Value::Bytes(v),
+            ObjectKey::Text(v) => Value::Text(v),
+        }
+    }
+}
```
NicolasDP: handy

plredmond:
Progress on vss-certificates.

So far I can't reconstruct the map because we decided to leave vss-key (that 35 byte public-key) raw. It's supposed to corresponds to the 28 byte stakeholder-id though.

plredmond:
vss-signing-key [64 bytes] is supposed to correspond to the stakeholder-id [28 bytes]

(see documentation)[https://github.com/input-output-hk/cardano-sl/blob/005076eb3434444a505c0fb150ea98e56e8bb3d9/core/src/Pos/Core/Ssc/VssCertificatesMap.hs#L36-L44]

blockchain/src/normal.rs
```diff
@@ -271,13 +272,19 @@ impl cbor::CborValue for VssCertificates {
...
 
 #[derive(Debug, Clone)]
 pub struct VssCertificate {
-    vss_key: cbor::Value, // TODO more than just cbor
-    expiry_epoch: types::EpochId,
-    signature: cbor::Value,   // TODO more than just cbor
-    signing_key: cbor::Value, // TODO more than just cbor
+    pub vss_key: cbor::Value, // 35 bytes public-key from http://hackage.haskell.org/package/pvss/docs/Crypto-SCRAPE.html#t:Point
+    pub expiry_epoch: types::EpochId,
+    pub signature: redeem::Signature,
```
NicolasDP: The size might be the same, but the meaning is completely different.
In wallet_crypto creates a new module named `vss.rs` and add the missing placeholder (no need to add the crypto yet).

wallet-cli/src/command/pretty.rs
```diff
@@ -3,32 +3,45 @@ use std::fmt;
...
+    //// actor ids
+    XPub(hdwallet::XPub),
     Stakeholder(address::StakeholderId),
 
     // recursive
-    List(Vec<Val>),
-    Tree(AST),
+    List(Vec<Val<'a>>),
+    Tree(AST<'a>),
+}
```
NicolasDP: The AST does not need to be **that** specific : one type for Hash (what ever algorithm) and one type for Signature is just as good.

```rust
pub enum Ast {
    // ...
    Hash(Vec<u8>),
    Signature(Vec<u8>),
    // ...
}
```

wallet-cli/src/command/pretty.rs
```diff
@@ -3,32 +3,45 @@ use std::fmt;
...
+use wallet_crypto::{address, cbor, config, hash, hdwallet, redeem, tx, util::hex};
 
 use ansi_term::Colour;
 
 // Constants for the fmt::Display instance
 static DISPLAY_INDENT_SIZE: usize = 4; // spaces
 static DISPLAY_INDENT_LEVEL: usize = 0; // beginning starts at zero
 
-type AST = Vec<(Key, Val)>;
+type AST<'a> = Vec<(Key<'a>, Val<'a>)>;
```
NicolasDP: While it is nice to optimise things by removing non necessary dynamic allocation, this is for pretty printing, it is okay if it takes a micro second for allocation. Keep this change, but don't pursue further, this is not the scope of this PR : we need block parsing/understanding for last week.

wallet-cli/src/command/pretty.rs
```diff
@@ -3,32 +3,45 @@ use std::fmt;
...
 // TODO: extend with blockchain-specific constructors with color
-pub enum Val {
+pub enum Val<'a> {
     // terminals
     Raw(String),
-    Hash(types::HeaderHash),
+    Hash(types::HeaderHash), // XXX: consider naming this with a more specific meaning, as we'll probably have other hashes?
+    //// numbers
     Epoch(u32),
     SlotId(u32),
```
NicolasDP: Just noticed these 2: Clearly here we want an `BlockDate` instead

blockchain/src/normal.rs
```diff
@@ -123,6 +123,189 @@ impl cbor::CborValue for Body {
...
+                    Ok(SscPayload::CertificatesPayload(vss))
+                }
+                _ => cbor::Result::array(array, cbor::Error::UnparsedValues),
+            }
+        })
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct Commitments(Vec<SignedCommitment>);
```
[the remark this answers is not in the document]
plredmond: `VssCertificateMap` and `CommitmentsMap` in haskell both drop their `StakeholderId` keys and get converted to `Array([..])` in the `cbor`. In the rust:

* `SignedCommitment.public_key` is supposed to correspond to a `StakeholderId` (from somewhere else in the block?)
* `VssCertificate.vss_key` is supposed to correspond to a `StakeholderId` (but we haven't decoded this one yet)

wallet-cli/src/command/pretty.rs
plredmond: the struct `normal::BlockSignature::{*}` includes two constructors which contain raw cbor and therefore aren't easily converted to `Signature(Vec<u8>)`.. I'd rather not convert them back to raw cbor bytes for pretty printing, so I'll use a dummy implementation which converts them to ascii bytes for now

wallet-cli/src/command/pretty.rs
plredmond: In the structure of the block we don't always get an `Epoch` together with a `SlotId`, so I think these need to stay separate.

NicolasDP: that's why I was talking about `BlockDate`, it is an enum that handles the Genesis and the Normal blocks separately.

wallet-crypto/src/redeem.rs
```diff
@@ -141,6 +141,8 @@ impl Signature {
         let bytes = hex::decode(hex)?;
         Self::from_slice(&bytes)
     }
+
+    pub fn to_bytes<'a>(&'a self) -> &'a [u8;SIGNATURE_SIZE] { &self.0 }
```
NicolasDP: :+1:

plredmond:
these are the blocks I'm testing the decoding & pretty-printing against
```shell
echo === four constructors of SccPayload
cargo run -- blockchain cat mainnet 35481e8ed589f800ebcbc596aaf06d0e5c55b8f6af9d8b134eef03bfdc198199
cargo run -- blockchain cat mainnet 65ce050df93f883f521ec91756393286d509886177d187c8f099e4fc3f0fb7cc
cargo run -- blockchain cat mainnet b05742df56b52c2fbdb9daba0a8d93bd9d1a8f8ff0185922f544b7234460bba6
cargo run -- blockchain cat mainnet 3d1f0f015452cb0266fdbbcc8f09a5dd144fd188f02049977314bc205ab5166f
echo == non empty vss in SccPayload
cargo run -- blockchain cat testnet f134770887ff970642f215bee0d5465e1fa8fa1f8c052c38709c42c69e1dcad5
cargo run -- blockchain cat testnet 629f3fc87e48af8739c94333eba2bbc8832e7f2ffb2688cd58cbfefade467b64
cargo run -- blockchain cat testnet 3e726761ee51a9cc0fafbe8108c9e460afb40f0ba9fd69005f8da5ea511ed073
```

wallet-crypto/src/redeem.rs
plredmond: This got removed in the following commit

wallet-cli/src/command/pretty.rs
plredmond: Ah, I see.. I'll make that change

PR #133, opened 2018-06-12
faster cbor serialisation/deserialisation

NicolasDP:
## some numbers:

| method | what | bench |
| ---------- | ------- | --------:|
| new way | block | 2,642 ns/iter (+/- 127) |
| old way | block | 68,539 ns/iter (+/- 336) |
| new way | address | 475 ns/iter (+/- 1) |
| old way | address | 2,717 ns/iter (+/- 16) |

also the command `ariadne wallet update <wallet-name>` takes now only 5sec per EPOCH while it used to need 30sec per EPOCH.

### Deprecating `cbor::Value`

The use of the enum `Value` which describes the different generic value one can find in cbor is now deprecated and we will need to strive to remove it:

- it is really not necessary in our case, we are encoding a specific protocol;
- it is expensive : memory and speed.

We are still using it in the `blockchain` for objects we don't know yet what they are. Genesis block is taking ~1.5sec to process, it is mainly using a recursive construction of this `Value`.

### Deserialisation: `RawCbor`

The old cbor parser used to allow any kind of [`Read`](https://doc.rust-lang.org/std/io/trait.Read.html) source. While this is interesting in a generic implementation, or in cases where we need to be able to receive streams of cbor from different kind of inputs. Here we only needed to be able to parse rather small objects: addresses (< 100 B), blocks (~650 B). They can all easily hold in memory.

The goal of this re-implementation was to:
- remove most of the dynamic allocation;
- go as fast as possible;
- separate the encoding/decoding.

So now the `RawCbor` is the object that allows us to deserialise cbor bytes into meaningful cardano objects. It is defined as follow:

```rust
pub struct RawCbor<'a>(&'a [u8]);
```

It is now only a pointer over an array of bytes. Every time we consume bytes, we move the pointer : there is no going backward.

### Serialisation

We now create an object `Serialiser` in which we can write cbor objects. The serialiser as the following property as well:

- it consumes itself in order to write anything, and returns itself if the write succeed;
- it can only go forward.

It is also 6 time faster to use this Serialisation object as it does not need to create the intermediate CBOR object.


vincenthz:
this is exactly what I wanted to see :100: good work

PR #155, opened 2018-07-06
clean up wallet implementations for bip44 and 2 level random indexes

NicolasDP:
The previous implementation tried to accommodate both models, however this made the code very unstable and more complicated. In this PR we try to define a unified interface for what a wallet should provide (the traits `Wallet` and `Account` in `cardano::wallet::scheme`) and we provide specific functions and behavior for the BIP44 Wallet and the 2 Level Random Indexes.


NicolasDP:
note: this will break compilation in the wasm bindings, I am currently working on the update. Users should not see changes in the APIs.

PR #158, opened 2018-07-10
C-bindings and address generator iterator

NicolasDP:
# C-Binding `cardano-c`

this is a small example of c-binding for use in other languages. The aim of this library is to
demonstrate a simple use case of C library using the `rust-cardano` crates.

So far it provides:

1. create a new wallet from a bip39 seed (64 bytes expected);
2. create new bip44 accounts from a wallet;
3. generate bip44 addresses from a wallet.

# Address iterator

while writing the example above, I found it a bit tedious to generate a list of addresses, the iterator is meant to facilitate the address generations:

```rust
let wallet = /*  */ ;
let account = wallet.create_account("Main", 0);

let addresses = account.address_generator(AddrType::External, 0)
    .take(20) // if not used, we may iterate up to 0x7FFFFFFF, the limit for soft derivation
    .enumerate()
    .filter(|(x, _)| x % 2 == 0)
    .drop(2)
    .map(|(_, xprv) { ExtendedAddr::new_simple(xprv.public()) })
    .collect::<Vec<_>>();
```

This will return 8 addresses, indexed: 4, 6 8, 10, 12, 14, 16 and 18.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
     "storage",
     "hermes",
     "cbor_event",
+    "cardano-c",
 ]
 [dependencies]
 printer = { path = "rust-crypto-wasm" }
--- a/cardano-c/Cargo.toml
+++ b/cardano-c/Cargo.toml
@@ -1,0 +1,7 @@
+[package]
+name = "cardano-c"
+version = "0.1.0"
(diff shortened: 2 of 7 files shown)


vincenthz:
a nice start !

PR #160, opened 2018-07-10
Simplify Hermes command line interface

### LINKED ISSUE #146: hermes simplified command line interface
link to #107 

hermes' command line interface is a bit messy and some options are missings:

We don't want commands anymore, only options:

- `--port <PORT>` the port number to listen to (default: `80`);
- `--blochain-root-dir <DIR>` the root directory where the blockchain**s** are (default `${HOME}/.hermes`);
- `--template <testnet|mainnet>` the template for the blockchain configuration. either `testnet` or `mainnet`, this is a cumulative option, can appear 0, 1 or 2 times:
    - no option passed: default is mainnet;
    - 1 option: get the name, sync with cardano protocol and serve it to http:
    - 2 options: get the names, make sure they are not equal, sync with cardano protocol and serve them to http.

To create the blockchain configuration file, it is possible to [reuse code from the wallet-cli](https://github.com/input-output-hk/rust-cardano/blob/master/wallet-cli/src/command/blockchain/mod.rs#L119-L134):

```rust
                let net_cfg = match value_t!(opts.value_of("template"), String).unwrap().as_str() {
                    "mainnet" => { net::Config::mainnet() },
                    "testnet" => { net::Config::testnet() },
                    _         => {
                        // we do not support custom template yet.
                        // in the mean while the error is handled by clap
                        // (possible_values)
                        panic!("invalid template option")
                    }
                };
                let config = resolv_network_by_name(&opts);
                let storage_config = config.get_storage_config();
                let _ = Storage::init(&storage_config).unwrap();

                let network_file = storage_config.get_config_file();
                net_cfg.to_file(&network_file)
```

edolstra:
This removes `hermes init` and allows all necessary configuration to be specified on the command line, e.g.
```
hermes start --port 4080 --networks-dir /var/lib/hermes/networks --template testnet
```

Fixes #146.

### DIFF
--- a/exe-common/src/network/native.rs
+++ b/exe-common/src/network/native.rs
@@ -33,7 +33,7 @@ impl PeerPool {
             match Connection::new(sockaddr, protocol_magic) {
                 Ok(connection) => connections.push(connection),
                 Err(Error::ConnectionTimedOut) => {
-                    warn!("connection peer `{}' address {} timedout, ignoring for now.", name, sockaddr)
+                    warn!("connection peer `{}' address {} timed out, ignoring for now.", name, sockaddr)
                 },
                 Err(err) => {
                     error!("connection peer `{}' address {} failed: {:?}", name, sockaddr, err);
--- a/hermes/src/config.rs
+++ b/hermes/src/config.rs
@@ -1,6 +1,6 @@
 use serde_yaml;
 
-use storage::{self, tmpfile::{TmpFile}};
+use storage::{self, Storage};
 use storage::config::StorageConfig;
 use exe_common::config::{net};
 use std::{io, result, path::{PathBuf, Path}, env::{VarError, self, home_dir}, fs};
@@ -55,16 +55,6 @@ impl Config {
         }
     }
 
-    pub fn open() -> Result<Self> {
-        let p = hermes_path()?.join("config.yml");
-        Self::from_file(p)
-    }
-
-    pub fn save(&self) -> Result<PathBuf> {
-        let p = hermes_path()?.join("config.yml");
-        self.to_file(p)
-    }
-
     pub fn get_networks_dir(&self) -> PathBuf { self.root_dir.clone() }
 
     pub fn get_networks(&self) -> Result<Networks> {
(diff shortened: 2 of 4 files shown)


exe-common/src/network/native.rs
```diff
@@ -33,7 +33,7 @@ impl PeerPool {
             match Connection::new(sockaddr, protocol_magic) {
                 Ok(connection) => connections.push(connection),
                 Err(Error::ConnectionTimedOut) => {
-                    warn!("connection peer `{}' address {} timedout, ignoring for now.", name, sockaddr)
+                    warn!("connection peer `{}' address {} timed out, ignoring for now.", name, sockaddr)
```
NicolasDP: cheers :D

hermes/src/main.rs
```diff
@@ -51,36 +53,47 @@ fn main() {
...
-            SubCommand::with_name("start")
-                .about("start explorer server")
+                .arg(Arg::with_name("TEMPLATE")
+                    .long("template")
+                    .takes_value(true)
+                    .value_name("TEMPLATE")
+                    .help("either 'mainnet' or 'testnet'; may be given multiple times")
+                    .required(false)
+                    .multiple(true)
+                    .default_value("mainnet")
```
NicolasDP: add `possible_values` too:

```rust
  .arg(Arg::with_name("TEMPLATE")
         // ...
        .possible_values(&["mainnet", "testnet"])
        .default_value("mainnet")
```

that way it is `clap` which is going to handle the error of unknown template input and will provide a more meaningful error message

edolstra:
It occurred to me that the configuration is still a bit stateful: it should really only use the networks specified with `--template`, but currently it will also use networks that were *previously* specified. So maybe we should remove `get_networks()` entirely and only use the specified networks.

edolstra:
Addressed the comments above.

hermes/src/config.rs
```diff
@@ -37,7 +38,8 @@ type Result<T> = result::Result<T, Error>;
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Config {
     pub root_dir: PathBuf,
-    pub port: u16
+    pub port: u16,
+    pub network_names: HashSet<String>
```
NicolasDP: I am not sure it is worth to use such structure here, I don't see a use case for more than 2 network types: mainnet and testnet (maybe staging?). So it won't be more than 2 or 3 networks here.

PR #169, opened 2018-07-26
Fix Typo In cardano.h

etscrivner:
Change `cardano_accont_delete` to `cardano_account_delete` to ensure that
header actually works when used to compile C programs. The previous typo was
preventing the header from being used as-is to build C programs.

### DIFF
--- a/cardano-c/cardano.h
+++ b/cardano-c/cardano.h
@@ -22,7 +22,7 @@ cardano_wallet *cardano_wallet_new_from_seed(const unsigned char * const seed_pt
 void cardano_wallet_delete(cardano_wallet *);
 
 cardano_account *cardano_account_create(cardano_wallet *wallet, const unsigned char *alias, unsigned int index);
-void cardano_accont_delete(cardano_account *account);
+void cardano_account_delete(cardano_account *account);
 
 unsigned long cardano_account_generate_addresses(cardano_account *account, int internal, unsigned int from_index, unsigned long num_indices, char *addresses_ptr[]);
 


etscrivner:
Used the following code to test this, would be good to clean up and perhaps work into an integration test?

```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include "cardano.h"

char* generate_single_address() {
  static const unsigned char* alias = "Test Wallet";
  static char* addresses[1];
  unsigned char bytes[64];
  FILE* fp;
  int i = 0;

  fp = fopen("/dev/urandom", "r");

  if (fp == NULL)
    return NULL;

  while (i < 64) {
    bytes[i] = (char)fgetc(fp);
    i++;
  }

  cardano_wallet *wallet = cardano_wallet_new_from_seed(
    bytes, 1
  );
  cardano_account *account = cardano_account_create(
    wallet, alias, 0);
  cardano_account_generate_addresses(
    account, 0, 0, 1, addresses);

  char* result = strdup(addresses[0]);

  cardano_account_delete(account);
  cardano_wallet_delete(wallet);

  return result;
}

int main(int argc, char* argv[]) {
  printf("Generating account...\n");
  char* result = generate_single_address();
  printf("Address: %s\n", result);
  free(result);
  return 0;
}
```

vincenthz:
thanks @etscrivner, just to a short heads up, we've really started the C rust library to demonstrate the capabilities, only 2 weeks ago (and the very rough .h last week), so it doesn't really export much and it's very lightly tested.

Putting regression tests (like the one you commented) and making sure the C library is solid and have everything properly exported is part of the plan, it's just catching us short, time wise.

Issue #153, opened 2018-07-05
exe common network sync code don't enshrine epoch pack


vincenthz:
epoch pack need to be made official (see CLI `epoch-from-pack`) after they are downloaded

NicolasDP:
closed by #171

Issue #80, opened 2018-05-21
Add support for block and epoch downloading from hermes


vincenthz:
through using hyper, add downloading epoch/block APIs

- [x] epoch download (done in #106)
- [x] block download
- [x] tip download

NicolasDP:
closed by #171

PR #171, opened 2018-07-27
net_sync() improvements

edolstra:
This adds the following improvements:

* It can synchronize up to the current tip, rather than just fetching entire epochs. Thus it will fetch individual blocks from the current epoch and store them as separate blobs on disk. When on a subsequent sync we have an entire epoch of blocks, they'll be packed into an epoch on disk.

* It respects the blockchain parameter 'k'. Thus it won't pack blocks into an epoch until we're at least k blocks into the next epoch.

* It maintains the `HEAD` tag to point to the most recently fetched tip. So it no longer relies on epochs to figure out from where to restart synchronization. In the future, we'll want a tip per peer (and probably a tag to point to the latest "stable" block).

* `ariadne blockchain sync` now has a flag `--native` to force it to sync using a native peer rather than a Hermes peer. This is useful for testing the native protocol.

* Unrelated, added for testing: a command `ariadne ls-blocks` to list the blocks in the blockchain.

Note that I changed `network::Api`: it no longer has a `fetch_epoch()` function. Instead, it has a `get_blocks()` function that fetches an arbitrary range of blocks. In the case of Hermes, this is optimized by fetching blocks in epochs, but that's an implementation detail that the caller doesn't need to know about. The reason for this change is that it simplifies `net_sync()` and `network::Api`, and it prevents the network layer from having a dependency on the storage layer.

Some limitations:

* It doesn't handle divergence yet: if the local tip is not an ancestor of the remote tip, it will barf. Obviously we'll need to handle this in the future (probably by restarting the sync from the latest stable block).

* Fetching the individual blocks in the current epoch from Hermes could be slow on a high-latency connection.

### DIFF
--- a/exe-common/src/config.rs
+++ b/exe-common/src/config.rs
@@ -192,7 +192,7 @@ pub mod net {
     pub struct Config {
         pub genesis: HeaderHash,
         pub genesis_prev: HeaderHash,
-        pub k: u32,
+        pub epoch_stability_depth: u32,
         pub protocol_magic: ProtocolMagic,
         pub epoch_start: EpochId,
         pub peers: Peers
@@ -205,7 +205,7 @@ pub mod net {
             Config {
                 genesis: HeaderHash::from_hex(&"89D9B5A5B8DDC8D7E5A6795E9774D97FAF1EFEA59B2CAF7EAF9F8C5B32059DF4").unwrap(),
                 genesis_prev: HeaderHash::from_hex(&"5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb").unwrap(),
-                k: 2160,
+                epoch_stability_depth: 2160,
                 protocol_magic: ProtocolMagic::default(),
                 epoch_start: 0,
                 peers: peers
@@ -219,7 +219,7 @@ pub mod net {
             Config {
                 genesis: HeaderHash::from_hex(&"B365F1BE6863B453F12B93E1810909B10C79A95EE44BF53414888513FE172C90").unwrap(),
                 genesis_prev: HeaderHash::from_hex(&"c6a004d3d178f600cd8caa10abbebe1549bef878f0665aea2903472d5abf7323").unwrap(),
-                k: 2160,
+                epoch_stability_depth: 2160,
                 protocol_magic: ProtocolMagic::new(633343913),
                 epoch_start: 0,
                 peers: peers
--- a/exe-common/src/config.rs
+++ b/exe-common/src/config.rs
@@ -192,6 +192,7 @@ pub mod net {
     pub struct Config {
         pub genesis: HeaderHash,
         pub genesis_prev: HeaderHash,
+        pub k: u32,
         pub protocol_magic: ProtocolMagic,
         pub epoch_start: EpochId,
         pub peers: Peers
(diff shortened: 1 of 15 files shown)


exe-common/src/config.rs
```diff
@@ -192,6 +192,7 @@ pub mod net {
     pub struct Config {
         pub genesis: HeaderHash,
         pub genesis_prev: HeaderHash,
+        pub k: u32,
```
vincenthz: I think we agreed on something a bit longer (my suggestion was something like 'epoch_stability_depth' IIRC)

exe-common/src/config.rs
```diff
@@ -204,6 +205,7 @@ pub mod net {
             Config {
                 genesis: HeaderHash::from_hex(&"89D9B5A5B8DDC8D7E5A6795E9774D97FAF1EFEA59B2CAF7EAF9F8C5B32059DF4").unwrap(),
                 genesis_prev: HeaderHash::from_hex(&"5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb").unwrap(),
+                k: 2160,
```
NicolasDP: at the top of this file, use a const to define this value:

```rust
const DEFAULT_EPOCH_STABILITY_DEPTH : usize = 2160;
```

also use `usize` here, so we can easily compare it with the number of hases in a collections, number of file in a directory, which are often typed as `usize`.

exe-common/src/network/api.rs
```diff
@@ -13,60 +10,19 @@ pub trait Api {
...
-    pub epoch_id: EpochId,
-    pub start_header_hash: HeaderHash,
-    pub previous_header_hash: HeaderHash,
-    pub upper_bound_hash: HeaderHash
+    /// Get the blocks in the half-open interval (from, to] (if
+    /// inclusive = false) or [from, to] (if inclusive = true). FIXME:
+    /// the inclusive = true case is only needed because the native
+    /// protocol doesn't support fetching from the genesis_prev hash.
+    fn get_blocks(&mut self, from: &BlockRef, inclusive: bool, to: &BlockRef,
+                   got_block: &mut FnMut(&HeaderHash, &Block, &RawBlock) -> ());
```
NicolasDP: use a where clause to simplify this function declaration:

```rust
fn get_blocks<F>(&mut self, from: &BlockRef, inclusive: bool, to: &BlockRef, got_block: &mut F)
    where F: FnMut(&HeaderHash, &Block, &RawBlock) -> ()
```

this will help reading the code a bit better.

exe-common/src/config.rs
edolstra: The reason for using a `u32` was to match the type of `SlotId`.

exe-common/src/config.rs
```diff
@@ -192,6 +192,7 @@ pub mod net {
     pub struct Config {
         pub genesis: HeaderHash,
         pub genesis_prev: HeaderHash,
+        pub epoch_stability_depth: u32,
```
NicolasDP: I still would like to see `usize` here, or to understand why this special `u32` type?

exe-common/src/network/api.rs
NicolasDP: why is there a mutable reference to this function ?
RawBlock = Block but serialised in cbor. What do you need the Block deserialised for?

edolstra: It passes a Block to the callee because the callee might need it (and indeed `net_sync()` does), and the caller already has it (since it performs some checks on the blocks and get the previous block in some cases), so it's more efficient to not have to deserialize it twice.

NicolasDP:
closes #153 
closes #80

PR #185, opened 2018-08-05
exe-common::sync: Hot-fix a u32 underflow panic

drozdziak1:
This was originally found on a first run of Hermes which would catch the
underflow on the genesis epoch. The finish_epoch() function would use an
imprecise logical expression to determine that the specified epoch has a
predecessor.

This commit fixes the underflow which would happen for
epoch_id equal to 0 a.k.a. genesis.

I don't think this is the correct way to handle this, because I believe
the sync module is inconsistent. It assumes in finish_epoch() that
`previous_epoch_id(N) == N - 1` always holds true and therefore that the
genesis epoch ID is always 0, while at the same time other parts of the
code use different methods than checking the ID to check for genesis
(e.g. `is_genesis()` on block date).

**What do you think should be the definitive way to determine the epoch
that's allowed not to have a predecessor?**

### DIFF
--- a/exe-common/src/sync.rs
+++ b/exe-common/src/sync.rs
@@ -234,7 +234,14 @@ fn finish_epoch(storage: &storage::Storage, epoch_writer_state: &mut EpochWriter
     tmpfile.render_permanent(&storage.config.get_index_filepath(&packhash)).unwrap();
     let epoch_time_elapsed = epoch_writer_state.write_start_time.elapsed().unwrap();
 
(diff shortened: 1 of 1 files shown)


drozdziak1:
One job errored in CI due to no output, I'm confident it doesn't have anything to do with my changes, could you please rerun the job when you get the chance?

edolstra:
Thanks, I messed up the assertion - `epoch_id > 0` was supposed to be `epoch_id == 0`.

Ideally, we would use `net_cfg.epoch_start` here. That will require passing `net_cfg` to `finish_epoch`.

PR #186, opened 2018-08-05
Fix all warnings in cardano-cli

drozdziak1:
This PR removes all warnings coming from cardano-cli. I didn't know what might
be your preference in terms of smashing particular warning kinds so I only did
this for the one crate.

**Note:** I made some arguable decisions like removing one unused method in
cardano-cli/src/blockchain/peer.rs or the unused stderr field from the Term
struct in cardano-cli/src/utils/term/mod.rs. I hope I connected the dots alright
on these two.

### DIFF
--- a/cardano-cli/src/blockchain/commands.rs
+++ b/cardano-cli/src/blockchain/commands.rs
@@ -1,8 +1,7 @@
 use std::path::PathBuf;
 use std::io::{Write};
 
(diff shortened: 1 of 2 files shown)


drozdziak1:
Got it, amended and force-pushed.

NicolasDP:
thanks

PR #202, opened 2018-08-08
Ntt cleanup

edolstra:
Some minor cleanups to the network code while I was working on the subscription code: 1) Fix `ack_to_syn` and `syn_to_ack`; and 2) turn `ControlHeader` into an explicitly numbered enum to remove the need for `control_header_{to,from}_u32`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -611,6 +611,22 @@ version = "0.1.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "num-derive"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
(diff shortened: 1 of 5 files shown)


NicolasDP:
I don't really see the value of adding these 2 crate dependencies to the project just to remove 2 small functions. I understand that it avoid redundancy of associating the enumeration to the correct values twice. I think it is preferable to keep the 2 functions. We aim to keep as little dependencies as possible. I think @vincenthz will agree with me on this.

Adding tests to check the the `from_value . to_value == id` would be great though. So we can guarantee that we did not made mistake when writing the converting functions.

NicolasDP:
somehow the last commit didn't update the PR or notify travis-ci.

edolstra:
The last commit is on another branch. I'll close this PR since the other branch contains this PR as well.

PR #204, opened 2018-08-13
Use the subscription mechanism to receive tip updates from the server

edolstra:
This allows the native server to push tip updates to the client, rather than the client having to poll the server for updates.

The most significant changes:

* `protocol::Connection` now sends a subscribe message at startup and handles incoming connections from the server. It sends an ACK to the server's SYN message. It saves all incoming data on the connection to `LightConnection::received`. (`LightConnection` is now used for both client and server connections.) When the server connection is closed, it calls `process_async_message()`. Currently, the only supported message type is `MsgHeaders`, which is handled by recording the new tip for subsequent calls to `get_tip` and `wait_for_new_tip`. 

* `struct Api` has a method `wait_for_new_tip()`. For the native backend, this blocks until a new tip is received from the server. For the Hermes backend, it polls periodically, sleeping in between.

* `net_sync()` has a new flag `sync_once`. When set to `false`, it will continuously synchronize to the latest tip obtained via `wait_for_new_tip()`. Otherwise it gives the old behaviour, i.e. doing a single sync.

* Hermes now starts a separate thread per configured network to run `net_sync()` in continuous mode.

### DIFF
--- a/cardano/src/block/block.rs
+++ b/cardano/src/block/block.rs
@@ -8,6 +8,7 @@ use cbor_event::{self, de::RawCbor};
 use super::types::{HeaderHash, SlotId, EpochId};
 use super::genesis;
 use super::normal;
+use super::super::cbor::hs::util::decode_sum_type;
 
 #[derive(Debug, Clone)]
 pub struct RawBlockHeaderMultiple(pub Vec<u8>);
@@ -305,15 +306,6 @@ impl cbor_event::de::Deserialize for BlockHeaders {
             }
         }
     }
-}
-
-fn decode_sum_type(raw: &mut RawCbor) -> cbor_event::Result<u64> {
-    let len = raw.array()?;
-    if len != cbor_event::Len::Len(2) {
-        return Err(cbor_event::Error::CustomError(
-            format!("Expected sum type but got array of {:?} elements", len)));
-    }
-    Ok(raw.unsigned_integer()?)
 }
 
 #[cfg(test)]
--- a/cardano/src/block/block.rs
+++ b/cardano/src/block/block.rs
@@ -49,6 +49,9 @@ pub enum BlockHeader {
     MainBlockHeader(normal::BlockHeader),
 }
 
+#[derive(Debug, Clone)]
+pub struct BlockHeaders(pub Vec<BlockHeader>);
+
 /// Block Date which is either an epoch id for a genesis block or a slot id for a normal block
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum BlockDate {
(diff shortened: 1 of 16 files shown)


cardano/src/block/block.rs
```diff
@@ -287,13 +281,41 @@ impl cbor_event::de::Deserialize for BlockHeader {
...
+            },
+            idx => {
+                Err(cbor_event::Error::CustomError(
+                    format!("Unsupported Headers: {}", idx)))
             }
         }
     }
 }
 
+fn decode_sum_type(raw: &mut RawCbor) -> cbor_event::Result<u64> {
```
NicolasDP: this function seems useful, could you push it down to `cardano/src/cbor/hs.rs` ?

edolstra: Done.

exe-common/src/sync.rs
```diff
@@ -173,6 +171,28 @@ pub fn net_sync<A: Api>(
     Ok(())
 }
 
+pub fn net_sync<A: Api>(
```
NicolasDP: can you document a bit this function ? There are lot of thing we need to remember when using this function, one of them being this function will use `thread::sleep` for a given amount of time when using `hermes`.

protocol/src/protocol.rs
```diff
@@ -352,60 +369,121 @@ impl<T: Write+Read> Connection<T> {
...
+        }
+    }
+
+    // Process a 'Headers' message.
+    pub fn process_async_headers(&mut self, msg: &[u8]) -> Result<()> {
+        let mut headers = cardano::block::BlockHeaders::deserialize(&mut RawCbor::from(msg))?;
+
+        info!("received {} asynchronous headers", headers.0.len());
+
+        if headers.0.len() > 0 {
```
NicolasDP:
```rust
if ! headers.is_empty() { /* ... */ }
```

protocol/src/protocol.rs
```diff
@@ -352,60 +369,121 @@ impl<T: Write+Read> Connection<T> {
...
+            self.latest_tip = headers.0.pop();
```
NicolasDP: I see you need to add the deref mut to `BlockHeaders`:

```rust
impl DerefMut for BlockHeaders {
    fn deref_mut(&mut self) -> &mut Self::Target { &mut self.0 }
}
```

protocol/src/protocol.rs
```diff
@@ -472,10 +472,10 @@ impl<T: Write+Read> Connection<T> {
     pub fn process_async_headers(&mut self, msg: &[u8]) -> Result<()> {
         let mut headers = cardano::block::BlockHeaders::deserialize(&mut RawCbor::from(msg))?;
 
-        info!("received {} asynchronous headers", headers.0.len());
+        info!("received {} asynchronous headers", headers.len());
 
-        if headers.0.len() > 0 {
-            self.latest_tip = headers.0.pop();
+        if let Some(latest_test) = headers.pop() {
```
NicolasDP: even better 👍

PR #205, opened 2018-08-14
Coin conversion

vincenthz:
* add a better display function
* add conversion to u64
* add conversion from u32


cardano/src/coin.rs
```diff
@@ -78,7 +78,7 @@ impl Coin {
 }
 impl fmt::Display for Coin {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self.0)
+        write!(f, "{}.{:06}", self.0 / 1000000, self.0 % 1000000)
```
NicolasDP: 👍

PR #210, opened 2018-08-20
Transaction sending support

edolstra:
Based on @NicolasDP's send-tx branch, rebased on master. It merges AnnounceTx and SendTx because the latter cannot be used without the former and splitting them creates issues like the lightweight connection not being closed properly on error.

Also some cleanup to protocol.rs. In particular Commands now return a protocol::Error rather than a string. This also gets rid of all the unwraps.


cardano/Cargo.toml
```diff
@@ -13,6 +13,7 @@ serde = "1.0"
 serde_derive = "1.0"
 cryptoxide = { path = "../cryptoxide" }
 cbor_event = { path = "../cbor_event" }
+log = "*"
```
NicolasDP: no lets remove this. we don't want the log in the cardano lib

vincenthz:
I don't understand why the Addressing api is now exposing index, looks a bit dangerous like this, but otherwise it's looking good !

Issue #89, opened 2018-05-21
more block validation


vincenthz:
Add some further blocks validation (header, body, proofs, etc)

We can see this as multi level validation (each back by their own function):

1. At the block level: check that the block is consistent with itself:
   * all hashes matches: e.g. body hash matches the one in the header, but other hashes also.
   * transactions semantic are correct: at least 1 input, at least 1 output, input sum matches output sum and fees, check fees according to a fee calculator
   * etc
2. At the epoch level: check that all the block within an epoch are valid:
   * signature verified
   * slot leaders keys (available in genesis block), matches each block signature.
3. Overall: check that utxo spent are valid.

edolstra:
I guess this can be closed? The only thing (AFAIK) that we're not validating is the slot leader selection, but that will soon be obsolete anyway.

vincenthz:
yep @edolstra

PR #232, opened 2018-08-28
Implement block consistency verification

edolstra:
This implements the ability to check the internal consistency of a block (level 1 of issue #89). In particular it checks:

* The various proofs in the block header.
* The signature of the block.
* The signatures of the VSS certificates.
* The transaction witnesses.
* Various other transaction checks such as uniqueness of inputs.

It adds a command `cardano-cli blockchain verify` to check every block in the chain, and `cardano-cli blockchain verify-block` to check a specific block.


cardano/src/tx.rs
```diff
@@ -354,11 +338,71 @@ impl cbor_event::de::Deserialize for Tx {
     }
 }
 
+/// A transaction witness is a vector of input witnesses
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
+pub struct TxWitness {
+    pub in_witnesses: Vec<TxInWitness>
+}
```
NicolasDP: Might be worth adding the following instead:

```rust
pub struct TxWitness(Vec<TxInWitness>);
pub Deref for TxWitness {
    type Taget = [TxInWitness];
    fn deref(&self) -> &Self::Target { self.0.deref() }
}
// add the AsRef, IntoIterator, FromIterator traits too
impl From<Vec<TxInWitness>> for TxWitness {
    fn from(witnesses: Vec<TxInWitness>) -> Self { TxWitness(witnesses) }
}
```

On the long run it will simplify our code, and makes the API more user friendly. This will allow construction with the following:

```rust
vec![...].into();
// or
[...].into_iter().collect()
```

cbor_event/src/de.rs
```diff
@@ -546,6 +560,11 @@ impl<'a> RawCbor<'a> {
         self.special()?.unwrap_bool()
     }
 
+    pub fn eof(&mut self) -> Result<()> {
+        if !self.0.is_empty() { return Err(Error::TrailingData) }
+        Ok(())
+    }
```
NicolasDP: I am not sure this function is needed. There is already an implementation for `Deref` which allows to access all the API of `[u8]`. So the user can already do:

```rust
let raw = RawCbor::from(&[]);

assert!(raw.is_empty()); // instead of raw.eof()
```

if you move `decode_complete` in this module, you could keep the useful error though:

```rust
impl<'a> RawCbor<'a> {
    /// just like `deserialize` but expects to consume all the data in the RawCbor
    /// (i.e. no trailing data).
    fn deserialize_complete<T>(&mut self) -> Result<T>
        where T: Deserialize
    {
        let v = self.deserialize()?;
        if ! self.is_empty() {
            Err(Error::TrailingData)
        } else { Ok(v) }
    }
}
```

cbor_event/src/se.rs
```diff
@@ -436,6 +441,11 @@ impl<W: Write+Sized> Serializer<W> {
         self.write_type(Type::Tag, tag)
     }
 
+    /// write a tag that indicates that the following list is a finite set
```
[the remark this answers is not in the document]
edolstra: According to a comment in `cardano-sl`, it will be standardized soon:
```
-- We stitch a `258` in from of a (Hash)Set, so that tools which
-- programmatically check for canonicity can recognise it from a normal
-- array. Why 258? This will be formalised pretty soon, but IANA allocated
-- 256...18446744073709551615 to "First come, first served":
-- https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml Currently `258` is
-- the first unassigned tag and as it requires 2 bytes to be encoded, it sounds
-- like the best fit.
```
I've added a reference to the IANA list.

cardano/src/block/normal.rs
```diff
@@ -411,18 +420,28 @@ impl VssCertificates {
...
+    // serialization of a map of StakeholderIds to VssCertificates
+    // (where StakeholderId is computed from each VssCertificate's
+    // signing key), rather than the serialization of a set of
+    // VssCertificates that's actually stored in the block.
+    pub fn serialize_for_proof<W: ::std::io::Write>(&self, serializer: cbor_event::se::Serializer<W>) -> cbor_event::Result<cbor_event::se::Serializer<W>> {
+        let mut hash = BTreeMap::<address::StakeholderId, &VssCertificate>::new();
+        for vss_cert in self.0.iter() {
+            hash.insert(address::StakeholderId::new(&vss_cert.signing_key), vss_cert);
+        };
+        cbor_event::se::serialize_fixed_map(hash.iter(), serializer)
```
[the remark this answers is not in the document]
edolstra: I looked into this and it turns out that the order of the VSSCertificates is actually supposed to be different:
* For computing the proof, items are sorted by StakeholderId (computed from `vss.signing_key`). 
* For serialization in the body, items are sorted in lexicographical order of the tuples `(vss.expiry_key, vss.vss_key, vss.signing_key, vss.signature)`.

For example, here are the VssCertificates from block `f134770887ff970642f215bee0d5465e1fa8fa1f8c052c38709c42c69e1dcad5` in staging:
```
[
  VssCertificate { 
    vss_key: PublicKey([88, 33, 2, 20, 99, 160, 53, 51, 153, 76, 206, 11, 203, 10, 172, 48, 76, 30, 38, 31, 65, 209, 98, 65, 37, 65, 76, 44, 87, 190, 230, 223, 119, 50, 65]), 
    expiry_epoch: 11, 
    signature: 4be2c4a968989245b8d3f459a0d543e882851e37e9361337115cb1b99dbce92cedb2acf71de47ae80a060323011268aa4f574a0b7cc86a5ac46787a3eb3a8f03, 
    signing_key: 2f19f1b86e694e5a23806ed30a34acaeb4131162297792e96a477f2659a1dc9d4de6f20f0252e7e56f16aa74938fde16f8e8636330f740de6f589a83c2a03c4b 
  }, 
  VssCertificate { 
    vss_key: PublicKey([88, 33, 2, 39, 113, 220, 242, 96, 59, 48, 98, 43, 74, 135, 222, 213, 102, 235, 137, 171, 113, 241, 251, 69, 155, 212, 1, 216, 105, 189, 71, 173, 104, 139, 179]), 
    expiry_epoch: 11, 
    signature: 50dc9439c6fc3f3b72e5f6847289eed2544acf52584bb02d634138722c1efb089890a063b8cc8c4b392a587599c73c8aab7bc9928f3082b2a906795791b6ca06, 
    signing_key: b1ce296b5a40ae25ceec8ccca41c8aa38706f741c2143e3ae4869ddbce2e0d8cf4e9d1c2a81c0c1d1ba445e43dd3e4c0fef4298f82f560d68b1c8ec34b467656 
  }, 
  VssCertificate { 
    vss_key: PublicKey([88, 33, 3, 98, 221, 212, 238, 242, 201, 168, 250, 115, 226, 175, 114, 58, 77, 164, 107, 236, 213, 251, 149, 100, 14, 222, 105, 129, 18, 183, 69, 103, 39, 116, 250]), 
    expiry_epoch: 11, 
    signature: f0f94a01befa445bad6667259aacc4b229effcfd9b59bd1a35ca45769025d24ec640957e2242544c41d14775264bcc191d44275b55c2dd9f8e90b0f8827c420b, 
    signing_key: 1d5e7e80a82861b1a8001df834c9080ef281aee96c4727c26664b4162ceb97b6d49c47ccb3dfc4caf2dc8036aa2dd7cd2cd7c5cc14a0011d5058d5800d0d068a }, 
  VssCertificate { 
    vss_key: PublicKey([88, 33, 3, 118, 141, 8, 109, 148, 115, 232, 219, 135, 251, 154, 41, 233, 135, 74, 188, 38, 197, 94, 17, 63, 226, 58, 186, 168, 247, 92, 118, 72, 70, 248, 243]), 
    expiry_epoch: 11, 
    signature: 6f12204bd79c436a96192d9ddbb1016e70fb52b8be0cf7c14911420e9603a52d3a46a2385b0b89716340a00deee6397e70869f3fd53825c994424097dbf7a100, 
    signing_key: d42e2f19d8078a5e7d82e91df183ecaad576a607da94be23ee9856b8a6848ac2e409e742bac1ad6ac90ec24d1f92212053c9157354c52a3ab14c4418744dc3a4 
  }, 
  VssCertificate { 
    vss_key: PublicKey([88, 33, 3, 131, 186, 22, 61, 120, 159, 207, 30, 90, 213, 87, 246, 157, 230, 244, 123, 41, 10, 225, 163, 119, 242, 89, 181, 248, 25, 185, 243, 203, 58, 155, 152]), 
    expiry_epoch: 11, 
    signature: 3e5c83e0d9f6699a47f0cf78ccd9582de59fbe0e993163beee332af5806afaaa486652eddfe84742563cf818bf77a31216fa5b45f81d0d621cc0ac9a67161403, 
     signing_key: ad0efdf334aa550ce7d78b5e340b4597a5cb1ea2a276d92c6344cf2c32924686fd249a6fb15eeed3b061a49aec2899ce833fa186e2f301a8451a0b81eada188a 
  }, 
  VssCertificate { 
    vss_key: PublicKey([88, 33, 3, 180, 216, 154, 19, 228, 205, 162, 229, 102, 250, 52, 246, 152, 128, 133, 69, 182, 192, 1, 187, 48, 117, 61, 243, 19, 3, 210, 176, 247, 251, 216, 102]), 
    expiry_epoch: 11, 
    signature: a4c8c2e9fee9d0a5bf10c16f028bed876c06dbb410e37f3f78797959c04244039bfb238309e21d3648d052ff959fd3c79bc90a1918cb1a951b7fa1516d1e6102, 
    signing_key: 1554fdf84c9b33960524a99866851ad216fa9991195e6b42861c70a534bf1a7f10788683ce9141abcf3d2174c2b7921d7b0f894fef5ffad897025b8094236d9e 
  }
]
```
So they're ordered by `vss_key` (since the `expire_epoch`s are all the same).

BTW, `BTreeMap` in Rust should indeed give the same order as the Haskell side, since it sorts by `StakeholderId`:
```
    encode = encodeMapSkel HM.size $ \f acc ->
        -- We need to encode the list with keys sorted in ascending order as
        -- that's the only representation we accept during decoding.
        foldr (uncurry f) acc . sortWith fst . HM.toList
```
and `StakeholderId` is a wrapper around `Basement.Block`, which compares lexicographically (hackage.haskell.org/package/basement-0.0.8/docs/src/Basement.Block.Base.html), just like `Ord` for `[u8]`.

cardano-cli/src/blockchain/commands.rs
```diff
@@ -343,7 +344,11 @@ pub fn cat( mut term: Term
         use utils::pretty::Pretty;
 
         let blk = rblk.decode().unwrap();
-        blk.pretty(&mut term, 0).unwrap();
+        if debug {
+            println!("{:?}", blk);
```
NicolasDP: use `{:#?}` instead of just `{:?}`. It will pretty print the rust structure over multilines and all. Easier to understand overall.

NicolasDP: and don't `println` use `writeln!(term, "{:#?}", blk)`. So we still have control on the output redirection. This will become handy when we need to _mock_ the output behavior into test.

PR #234, opened 2018-09-04
 Add magic + version number to all binary files

### LINKED ISSUE #230: Version all stored data
add magic header and version to all files, to prevent difficult upgrade issue in the future.

edolstra:
The header format is a 16-byte structure, consisting of the magic string `\xfeCARDANO`, a 4-byte file type identifier (e.g. `EPOC` for an epoch pack) and a 4-byte version number (currently 1 for all types).

Fixes  #230.

Warning: this breaks all current binary files.

### DIFF
--- a/storage/src/lib.rs
+++ b/storage/src/lib.rs
@@ -14,6 +14,7 @@ pub mod epoch;
 pub mod refpack;
 pub mod utils;
 pub mod containers;
+pub mod magic;
 use std::{fs, io, result};
 
 pub use config::StorageConfig;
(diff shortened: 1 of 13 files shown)


storage/src/lib.rs
```diff
@@ -154,13 +154,7 @@ pub mod blob {
...
 
     pub fn read(storage: &super::Storage, hash: &super::BlockHash) -> Result<RawBlock> {
-        let mut content = Vec::new();
-        let path = storage.config.get_blob_filepath(&hash);
-
-        let mut file = fs::File::open(path)?;
-        file.read_to_end(&mut content)?;
-
-        Ok(RawBlock::from_dat(content))
+        Ok(RawBlock::from_dat(self::read_raw(storage, hash)?))
```
NicolasDP: 👍

PR #247, opened 2018-09-11
Add function for verifying blocks in the context of the chain

edolstra:
Currently, this only verifies slot leaders and block dates.

### DIFF
--- a/cardano/src/block/verify_chain.rs
+++ b/cardano/src/block/verify_chain.rs
@@ -1,0 +1,104 @@
+use block::*;
+use address;
+use config::{ProtocolMagic};
+
+pub struct ChainState {
+    pub prev_block: HeaderHash,
+    pub prev_date: Option<BlockDate>,
+    pub slot_leaders: Vec<address::StakeholderId>,
+}
+
+impl ChainState {
+    pub fn new(genesis_prev: &HeaderHash) -> Self {
+        ChainState {
+            prev_block: genesis_prev.clone(),
+            prev_date: None,
+            slot_leaders: vec![]
+        }
+    }
+}
+
+pub fn verify_block_in_chain(
+    protocol_magic: ProtocolMagic,
+    chain_state: &mut ChainState,
+    block_hash: &HeaderHash,
+    blk: &Block) -> Result<(), Error>
+{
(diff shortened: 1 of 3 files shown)


cardano/src/block/verify_chain.rs
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+}
+
+pub fn verify_block_in_chain(
+    protocol_magic: ProtocolMagic,
+    chain_state: &mut ChainState,
+    block_hash: &HeaderHash,
+    blk: &Block) -> Result<(), Error>
+{
```
NicolasDP: why not make this function a method of `ChainState`?

```rust
    pub fn verify_block(&mut self, &HeaderHash, &Block) -> Result<&mut Self, Error>;
```

cardano/src/block/verify_chain.rs
```diff
@@ -0,0 +1,104 @@
+use block::*;
+use address;
+use config::{ProtocolMagic};
+
+pub struct ChainState {
+    pub prev_block: HeaderHash,
+    pub prev_date: Option<BlockDate>,
+    pub slot_leaders: Vec<address::StakeholderId>,
```
NicolasDP: I would also add the expected `protocol_magic` here as it is a parameter of the blockchain and it must be consistent within the whole blockchain.

cardano/src/block/verify_chain.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    res
+}
+
+fn do_verify(
+    protocol_magic: ProtocolMagic,
+    chain_state: &mut ChainState,
+    block_hash: &HeaderHash,
+    blk: &Block) -> Result<(), Error>
+{
```
NicolasDP: you can also make this function a private member of the ChainState:

```rust
    fn do_verify(&mut self, &HeaderHash, &Block) -> Result<&mut Self, Error>;
```

cardano/src/block/verify_chain.rs
NicolasDP: The return of `&mut Self` is not necessary here.

edolstra:
@NicolasDP Updated.

PR #248, opened 2018-09-11
Remove serde dependency

### LINKED ISSUE #237: remove serde as a dependency of the cardano crate
This is in order to save some space for our low memory targets.

This will certainly affect:

1. wasm-cardano-js;
2. cardano-cli;

as they might be using some of the cardano object's serde implementations.

NicolasDP:
closes #237 

### Before merging:

- [ ] update cardano-cli to point to this branch with the appropriate changes


NicolasDP:
will use #256 instead

PR #267, opened 2018-09-18
Use the block streaming API for synchronization

edolstra:
This is both significantly faster and simpler than repeatedly asking for headers and corresponding blocks.


protocol/src/protocol.rs
```diff
@@ -561,6 +564,55 @@ pub mod command {
...
+            // TODO: we may want to update the window size dynamically.
+            window_left -= 1;
+            if window_left < window_size / 2 {
+                window_left = window_size;
+                // Note: we don't prepend MsgStream here.
+                connection.send_bytes(id, &packet::send_msg_stream_update(window_left).1)?;
+            }
+        }
+
+        // FIXME
```
NicolasDP: what does the FIXME stand for here?

PR #271, opened 2018-09-25
Calculate transaction fee

datnv-ibl:
Add a function to calculate transaction fee, return zero if failed.

### DIFF
--- a/cardano/src/coin.rs
+++ b/cardano/src/coin.rs
@@ -100,6 +100,10 @@ impl Coin {
             Ordering::Less    => CoinDiff::Negative(Coin(against.0 - self.0)),
         }
     }
+
+    pub fn to_integral(self) -> u64 {
+        return self.0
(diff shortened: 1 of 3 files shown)


cardano/src/coin.rs
```diff
@@ -100,6 +100,10 @@ impl Coin {
             Ordering::Less    => CoinDiff::Negative(Coin(against.0 - self.0)),
         }
     }
+
+    pub fn to_integral(self) -> u64 {
+        return self.0
+    }
```
NicolasDP: This function is not needed because of the presence of `Deref` with the `Target = u64`.

datnv-ibl:
Thank you, I have changed the c function to use uint64_t and remove to_integral.

NicolasDP:
Thanks, I'll merge it as soon as it is green on the CI.

PR #277, opened 2018-09-27
Support testnet

edolstra:
This updates the genesis hashes and adds support for non-AVVM balances and the NetworkMagic field in addresses.

It also improves get_block error handling a bit.


cardano/src/config.rs
```diff
@@ -62,6 +63,21 @@ impl cbor_event::Deserialize for ProtocolMagic {
     }
 }
 
+pub type NetworkMagic = Option<i32>;
```
NicolasDP: I would rather like to use a specific type for this:

```rust
pub struct NetworkMagic {
    NoneRequired,
    Expected(i32)
}
```

cardano/src/config.rs
```diff
@@ -62,6 +63,21 @@ impl cbor_event::Deserialize for ProtocolMagic {
...
+        // FIXME: is there a better way to determine whether to emit
+        // NetworkMagic? There is a requiresNetworkMagic field in
+        // lib/configuration.yaml, but not in the genesis data.
+        if pm == ProtocolMagic::default() || *pm == 633343913 {
+            None
+        } else {
+            Some(*pm as i32)
+        }
+    }
+}
```
NicolasDP: I think it would be easier to put it, in the configuration of the CLI and hermes.

cardano/src/address.rs
```diff
@@ -181,25 +182,30 @@ impl cbor_event::de::Deserialize for StakeDistribution {
 #[cfg_attr(feature = "generic-serialization", derive(Serialize, Deserialize))]
 pub struct Attributes {
     pub derivation_path: Option<HDAddressPayload>,
-    pub stake_distribution: StakeDistribution
+    pub stake_distribution: StakeDistribution,
+    pub network_magic: config::NetworkMagic,
     // attr_remains ? whatever...
 }
 impl Attributes {
     pub fn new_bootstrap_era(hdap: Option<HDAddressPayload>) -> Self {
```
NicolasDP: I think we need to update the parameter of these functions to add their `NetworkMagic`. Or at least to deprecate them (or enable them only for the tests maybe?) so we don't have to do too much rewrite everywhere...

NicolasDP:
I guess the golden tests for the new address are somewhere there : https://github.com/input-output-hk/cardano-sl/tree/release/1.3.1/core/test/golden but the output data is not readable... we will need to do some digging

edolstra:
Some updates:

* `NetworkMagic` is now an enum rather than `Option<i32>`.

* The `Attributes` and `ExtendedAddr` constructors (and various address generating functions) now require a `NetworkMagic` argument. `cardano-cli` will need to be updated for this.

* Added a test for encoding/decoding address with `NetworkMagic`.

cardano-c/src/address.rs
```diff
@@ -36,9 +36,9 @@ pub extern "C" fn cardano_address_is_valid(c_address: *mut c_char) -> c_int {
 }
 
 #[no_mangle]
-pub extern "C" fn cardano_address_new_from_pubkey(c_xpubkey: XPubPtr) -> AddressPtr {
+pub extern "C" fn cardano_address_new_from_pubkey(c_xpubkey: XPubPtr, protocol_magic: ProtocolMagic) -> AddressPtr {
     let xpub = unsafe { c_xpubkey.as_ref() }.expect("Not a NULL PTR");
-    let ea = ExtendedAddr::new_simple(xpub.clone());
+    let ea = ExtendedAddr::new_simple(xpub.clone(), NetworkMagic::from(protocol_magic));
```
NicolasDP: no need for this PR. just out of curiosity or knowledge sharing. Did you know you could use `Into` for everytime you have `From` implemented ?

This allows you to write things like:

```rust
    let ea = ExtendedAddr::new_simple(xpub.clone(), protocol_magic.into());
```

I like the verbose way, less chance to get it wrong. Still less character too.

cardano/src/address.rs
```diff
@@ -249,18 +262,27 @@ impl cbor_event::de::Deserialize for Attributes {
...
-                1 => derivation_path    = Some(cbor_event::de::Deserialize::deserialize(raw)?),
+                ATTRIBUTE_NAME_TAG_STAKE =>
+                    stake_distribution = raw.deserialize()?,
+                ATTRIBUTE_NAME_TAG_DERIVATION =>
+                    derivation_path    = Some(raw.deserialize()?),
+                ATTRIBUTE_NAME_TAG_NETWORK_MAGIC => {
+                    // Yes, this is an integer encoded as CBOR encoded as Bytes in CBOR.
+                    let bytes = raw.bytes()?;
+                    let n = RawCbor::from(bytes.bytes()).deserialize::<u32>()?;
+                    network_magic = NetworkMagic::Magic(n);
```
NicolasDP: haskell :weary:

sjmackenzie:
We're looking for test net functionality, could you be a dear and merge this?

edolstra:
Synced with master. @NicolasDP @vincenthz Want to merge?

NicolasDP:
@edolstra I cannot merge. This branch does not compile

sjmackenzie:
great chaps, thanks

PR #282, opened 2018-10-03
Update input selection policy

NicolasDP:
The idea is to provide slightly better input selection algorithm than the current one.

here we provide mainly 2 type of input selection:

* large input first, this is not ideal but it will have the advantage to be rather fast;
* blackjack: just like large input first, but selecting only the inputs that are below the requested output value.


cardano/src/fee.rs
```diff
@@ -107,11 +107,20 @@ impl LinearFee {
...
 /// Calculation of fees for a specific chosen algorithm
 pub trait FeeAlgorithm {
+    fn estimate_overhead(&self, num_bytes: usize) -> Result<Option<Fee>>;
+
     fn calculate_for_txaux(&self, txaux: &TxAux) -> Result<Fee>;
     fn calculate_for_txaux_component(&self, tx: &Tx, witnesses: &Vec<TxInWitness>) -> Result<Fee>;
 }
 
 impl FeeAlgorithm for LinearFee {
+    fn estimate_overhead(&self, num_bytes: usize) -> Result<Option<Fee>> {
```
vincenthz: need to change the name, it doesn't really do the function justice

vincenthz:
💯

Issue #31, opened 2018-04-09
release 1.2 to nodejs.com


NicolasDP:
when Milestone 2 is finished, we can publish a new release of the JS glue to nodejs.com

PR #285, opened 2018-10-06
Adding Storage::load function for fixing issues in cardano-cli

ilap:
Cardano-cli has some issues see [#31](https://github.com/input-output-hk/cardano-cli/issues/31).
This PR and replacing the Storage::init /w the new Storage::load function in the [cardano-cli's Blockchain::load](https://github.com/input-output-hk/cardano-cli/blob/master/src/blockchain/mod.rs#L96), would fix the issues.

### DIFF
--- a/storage/src/lib.rs
+++ b/storage/src/lib.rs
@@ -91,6 +91,26 @@ pub struct Storage {
 }
 
 impl Storage {
+    // The load function should not create any dirs.
+    pub fn load(cfg: &StorageConfig) -> Result<Self> {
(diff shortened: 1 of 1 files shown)


vincenthz:
that looks fine in principle, but we need to move carefully between the current api which is overloading initialization and loading for the purpose of making sure we don't accidently use the wrong thing.

we really need to keep the current api and put a deprecation on it, and then offer an proper `initialization` that will fail if things already exist and a load that will fail is things not exist.

PR #286, opened 2018-10-07
Utxo storage

edolstra:
`net_sync()` now verifies blocks and writes the utxo state at the end of each epoch to disk to `epoch/<N>/utxos`.  Rather than storing a full dump of the utxo state at the start of each epoch, we store deltas relative to some parent epoch, where the parent is selected such that each full chain has O(lg epoch) deltas. This    reduces storage from 5.1 GiB to 356 MiB for the current mainnet (73 epochs).

I moved the genesis data from `cardano-cli` to `exe-common`, since `cardano-http-bridge` now needs it to sync.


vincenthz:
does that means that a full utxo dump is roughly ~70mb ?

storage/src/utxo.rs
```diff
@@ -0,0 +1,174 @@
...
+/// keys/values that are in 'm2' but not in 'm1'.
+fn diff_maps<'a, K, V>(m1: &'a BTreeMap<K, V>, m2: &'a BTreeMap<K, V>) -> (BTreeSet<&'a K>, BTreeMap<&'a K, &'a V>)
+    where K: Ord
+{
+
+    let mut removed = BTreeSet::new();
+    let mut added = BTreeMap::new();
+
+    let mut i1 = m1.iter();
+    let mut i2 = m2.iter();
```
NicolasDP: what about `zip` the 2 iterators? I think it will simplify your loop below

storage/src/config.rs
```diff
@@ -27,6 +27,7 @@ impl StorageConfig {
             StorageFileType::Blob => p.push("blob/"),
             StorageFileType::Tag => p.push("tag/"),
             StorageFileType::Epoch => p.push("epoch/"),
+            StorageFileType::Utxo => panic!("utxos don't have a directory"),
```
vincenthz: isn't that `p.push("utxos")` ?

vincenthz:
This is fine, apart from maybe the epoch storage api. I think it would be helpful to cut the two apis, because otherwise we don't have the ability to not create utxo specially on the client, I think the utxo database(s) is only going to be present in hermes and the full nodes, I'm less certain we want the same handling on the client.

vincenthz:
furthermore, it might be useful to distinguish utxo / utxodiff storage files and offer some simple api to create and apply/rollback diffs ? For example a client could store the latest stable utxo database,
apply a utxodiff, and store the new utxo database along with a reverse utxodiff (e.g. patch -R).

edolstra:
I've generalized the API to allow generation of deltas between arbitrary epochs as well as full utxo dumps. `net_sync()` currently use the logarithmic-length chain of deltas, but this can be changed to full dumps or something else without changing the file format.

edolstra:
Also `epoch_create` no longer needs to write utxo states.

vincenthz:
👍

PR #287, opened 2018-10-09
Make blackjack a bit random so we have a proper implementation of the algo

NicolasDP:


cardano/src/input_selection/simple_selections.rs
```diff
@@ -495,7 +524,7 @@ mod unit_tests {
         let inputs = vec![input1, input2, input3.clone(), input4.clone()];
         let outputs = vec![output1];
 
-        let selected = vec![input3, input4];
+        let selected = vec![input4, input3];
```
NicolasDP: here I only changed the order because of the randomnisation of the algorithm

PR #290, opened 2018-10-10
Checked address

vincenthz:
* rename Addr to HashedSpendingData
* introduce a new type Addr to be able to keep the address in binary form but validated. this is the equivalent to the result of the cbor serialization of ExtendedAddr

### DIFF
--- a/cardano/src/address.rs
+++ b/cardano/src/address.rs
@@ -364,7 +364,8 @@ impl fmt::Display for Addr {
 
 impl cbor_event::se::Serialize for Addr {
     fn serialize<W: ::std::io::Write>(&self, serializer: Serializer<W>) -> cbor_event::Result<Serializer<W>> {
-        serializer.write_bytes(self.as_ref())
+        let raw_cbor = RawCbor::from(&self.0);
+        serializer.append_raw_cbor(&raw_cbor)
     }
 }
 impl cbor_event::de::Deserialize for Addr {
--- a/cardano/src/address.rs
+++ b/cardano/src/address.rs
@@ -361,6 +361,18 @@ impl fmt::Display for Addr {
     }
 }
 
+impl cbor_event::se::Serialize for Addr {
+    fn serialize<W: ::std::io::Write>(&self, serializer: Serializer<W>) -> cbor_event::Result<Serializer<W>> {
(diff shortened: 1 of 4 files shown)


vincenthz:
only cosmetically, should be dealt by git automatically. it's only changing the attributes in #277

cardano/src/address.rs
```diff
@@ -1,4 +1,13 @@
 //! Address creation and parsing
+//!
+//! Address components are:
+//! * `HashedSpendingData` computed from `SpendingData`
```
NicolasDP: computed from `SpendingData` **and `Attributes` and `AddrType`**

cardano/src/address.rs
```diff
@@ -333,6 +456,10 @@ impl ExtendedAddr {
         ExtendedAddr::new(AddrType::ATPubKey, SpendingData::PubKeyASD(xpub), Attributes::new_bootstrap_era(None))
     }
 
+    pub fn to_address(&self) -> Addr {
+        Addr(cbor!(self).unwrap()) // unwrap should never fail from strongly typed extended addr to addr
```
NicolasDP: it is the cbor serialization that can fails. But if it fails here it means the user does not have anymore memory to _extend_ the `Vec` used in `cbor_event` to write the bytes of the cbor encoded address.

cardano/src/address.rs
```diff
@@ -257,72 +266,186 @@ impl cbor_event::de::Deserialize for Attributes {
...
+impl fmt::Display for Addr {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", base58::encode(&self.0))
+    }
+}
+
+impl cbor_event::se::Serialize for Addr {
+    fn serialize<W: ::std::io::Write>(&self, serializer: Serializer<W>) -> cbor_event::Result<Serializer<W>> {
+        serializer.write_bytes(self.as_ref())
+    }
```
NicolasDP: I don't think this is what you want to write here. Here `write_bytes` will write the CBOR byte object. It will start with the length and then the bytes of the `Vec`. It won't be a valid ExtendedAddr anymore after that.

PR #305, opened 2018-10-28
nix: build using carnix

sjmackenzie:
Carnix is fast becoming the standard way of doing rust in nixpkgs.


edolstra:
This is probably not desirable at this time since most contributors don't use Nix, so they won't be able to regenerate `Cargo.nix` when needed.

sjmackenzie:
okay closing.

PR #310, opened 2018-11-01
expect_bytes(): Only decode ACK/SYN once per light connection

edolstra:
Fixes

```
thread 'tokio-runtime-worker-1' panicked at 'assertion failed: self.remaining() >= dst.len()', /home/eelco/.cargo/registry/src/github.com-1ecc6299db9ec823/bytes-0.4.10/src/buf/buf.rs:241:9
```

This happened because expect_bytes() tried to decode the incoming data as an ACK/SYN for every block of incoming data, but some of them might be too small. Really only the check
```
if bytes.len() != 9 { return None }
```
is necessary, but it seems like a good idea to make sure we only receive one ACK/SYN per connection. Maybe connection data might be confused for an ACK/SYN message (i.e. if it starts with 0x53 or 0x41).

### DIFF
--- a/protocol-tokio/src/protocol/codec/message.rs
+++ b/protocol-tokio/src/protocol/codec/message.rs
@@ -178,6 +178,7 @@ impl Message {
 
 fn decode_node_ack_or_syn(lwcid: nt::LightWeightConnectionId, bytes: &Bytes) -> Option<Message> {
     use bytes::{Buf, IntoBuf};
+    if bytes.len() != 9 { return None }
     let mut buf = bytes.into_buf();
     let key = buf.get_u8();
     let v = buf.get_u64_be();


protocol-tokio/src/protocol/inbound_stream.rs
```diff
@@ -58,6 +58,7 @@ pub enum Inbound {
 pub struct InboundStream<T> {
     stream: SplitStream<nt::Connection<T>>,
     state: Arc<Mutex<ConnectionState>>,
+    node_received: bool,
```
NicolasDP: I don't see the value of checking this. We don't need this at all because if the next message ends up in the `InboundStream::forward_message` it will fail to find the node_id and will return an error. The need for the `node_received` is already enforced by design of the protocol itself.

PR #314, opened 2018-11-05
Fix build error on rust 1.27

yihuang:
https://github.com/rust-lang/rust/issues/49502

### DIFF
--- a/exe-common/src/parse_genesis_data.rs
+++ b/exe-common/src/parse_genesis_data.rs
@@ -3,7 +3,7 @@ use std::io::Read;
 use serde_json;
 use cardano::{config, fee, block, coin, redeem};
 use base64;
-use std::time::{Duration, SystemTime};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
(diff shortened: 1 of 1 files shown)


NicolasDP:
thanks for looking into it. Can you tell us more why you need *this* specific version of the rust compiler ?

My concern is that `SystemTime::UNIX_EPOCH` is a recent change and that it might deprecates the former way to get this value.

Also, there is a [serious security concern regarding all compilers prior 1.29](https://blog.rust-lang.org/2018/09/25/Rust-1.29.1.html) and I can only recommend you update the compiler version you depend on.

mzabaluev:
From [discussion](https://github.com/rust-lang/rust/issues/49502) on the library change, the issue was more about discoverability of the constant, and the old name is OK otherwise. In any case, it has to stay there for Rust 1.0 timeframe.

That said, we may prefer to only target new compiler versions in the code, and switch to the 2018 edition as the baseline when it gets released.

yihuang:
It doesn't have the problem with the recent version of rust, my version was too old.

PR #317, opened 2018-11-05
add function to subscribe and keep_alive

NicolasDP:


protocol-tokio/src/protocol/codec/message.rs
```diff
@@ -123,6 +126,10 @@ impl Message {
                 Data(lwcid, MessageType::MsgGetBlocks.encode_with(&gb))
             }
             Message::Block(lwcid, b) => Data(lwcid, MessageType::MsgBlock.encode_with(&b)),
+            Message::Subscribe(lwcid, keep_alive) => {
+                let keep_alive : u64 = if keep_alive { 43 } else { 42 };
```
mzabaluev: Is there other "source of truth" for these magic numbers other than this code?

protocol-tokio/src/protocol/codec/message.rs
```diff
@@ -171,6 +178,11 @@ impl Message {
             MessageType::MsgBlock => {
                 Ok(Message::Block(lwcid, cbor.deserialize_complete().unwrap()))
             }
+            MessageType::MsgSubscribe1 => {
+                let v : u64 = cbor.deserialize_complete().unwrap();
+                let keep_alive = v == 43;
```
mzabaluev: This silently treats all values other than 43 or 42 as "non-keepalive". I hope it's OK.

protocol-tokio/src/protocol/codec/message.rs
NicolasDP: my theory is that the dev behind it chose 42 for false, and then followed up with 43 for true... or maybe it is because 6 by 9... unfortunately it is not easy to come by more regarding these...

PR #328, opened 2018-11-08
storage: Improve iterator API

mzabaluev:
The iterators in crate `storage` could use some improvement.

* The `Iter` type does not seem to get much use, is internally branchy and does not have a clearly-defined meaning. It is better to introduce purpose-tailored iterators in the vein of `Range` and `ReverseIter`, as the use cases for them become apparent.
* Explicit iterator constructors are less ergonomic than iterator-returning methods on the iteratee object.
* The `Result` return type of the constructors may be surprising, as the `new` constructor convention is to return `Self`.
* `ReverseIter` names its constructor `from`, which should be avoided for ad-hoc conversion methods as it may get ambiguous with the blanket `From` impl in some generic code:

    ```rust
    fn from_a<T>(a: T)
    where
        ReverseIter: From<T>,
    {
        let _ = ReverseIter::from(a);  // breaks here
    }
    ```

The changes deprecate `Iter` and the constructor methods, and add a `reverse_from` method on `Storage` as a more convenient replacement similar to the already existing `range`.


mzabaluev:
> @mzabaluev , please merge _if_ you are confident the the CLI will still work after.

It will, because the CLI build is pinned to a particular commit in rust-cardano. Also, the CLI uses its own copy of `Iter`.

PR #336, opened 2018-11-14
storage: Improved block location API

mzabaluev:
The `Loose` variant of `BlockLocation` can carry the hash that was sought.
This removes a redundant, error-prone parameter of `block_read_location`.

Free functions in crate `storage` that have `Storage` as the first
parameter are replaced with methods on `Storage`. The existing functions
are marked deprecated:

- `block_location`
- `block_read_location`
- `block_read`

`BlockLocation` is also deprecated in favor of `BlockLocation2`, which will
be renamed back to `BlockLocation` in the next breaking release.


vincenthz:
we don't need to assure compatibility just yet, `BlockLocation2` is not really needed at this stage. we can just change BlockLocation directly for now.

Issue #329, opened 2018-11-08
Deref considered harmful


mzabaluev:
The crates of this project liberally declare `Deref` impls to facilitate coercion between various types. This is [problematic][c-deref] for API maintainability, as it makes the method set available for a type harder to reason about, and makes the result type of operator `*` non-evident from the code. `Deref` should be reserved for smart pointers, and tightly coupled types like owned and borrowed variants for the same data. For "subclass" type relationships and easy conversions to a more basic underlying type, it's preferable to use `AsRef` and explicit ad-hoc reference conversion methods.

[c-deref]: https://rust-lang-nursery.github.io/api-guidelines/predictability.html#c-deref

vincenthz:
I agree, I had a conversation about this topic a while ago with @NicolasDP, and we settled that it was breaking down abstractions to do this. So we should not use this kind of things in the future (apart from exceptions), and eventually remove the one we already have when we clean things up.

it's way more useful for limiting types usage, to set up only the apis you want with strong types, even if they look like other api, than using `Deref`

mzabaluev:
The `Deref` impls for hash types slicing them down to a `[u8; HASH_SIZE]` are particularly problematic, as this can occur without any visible hint in a deref coercion.

PR #338, opened 2018-11-15
Conditionally disable Deref impls for hash types

mzabaluev:
Automatic dereferencing from hash newtypes to their underlying byte
arrays is fraught with possibility of overlooked coercions and method
API drift.

As the first step, introduce a feature "bad-deref", enabled by default,
to conditionally disable the `Deref` impls for hash types and make sure
the libraries compile without the feature. Provide an ad-hoc
reference conversion method `.as_hash_bytes()` on hash types to access
the byte array in a more explicit fashion.

### DIFF
--- a/cardano/Cargo.toml
+++ b/cardano/Cargo.toml
@@ -28,6 +28,7 @@ unicode-normalization = "0.1"
 quickcheck = "0.7"
 
 [features]
-default = []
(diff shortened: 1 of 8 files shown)


mzabaluev:
This is one step to resolving #329.

PR #346, opened 2018-12-04
Fixed comment over `TxBuilder.balance`

vantuz-subhuman:
Original comment for `Positive` case is wrong ("more outputs+fees than inputs"), and also a bit hard to read the logic. Tried to make it as obvious as possible.


NicolasDP:
thanks

PR #352, opened 2018-12-18
Make diff_maps() public

edolstra:
I need this function in jormungandr.


storage/src/chain_state.rs
```diff
@@ -233,7 +233,7 @@ fn parent_for_epoch(epoch: EpochId) -> Option<EpochId> {
 /// Compute the diff from BTreeMap 'm1' to BTreeMap 'm2', returning
 /// the set of keys in 'm1' that are not in 'm2', and the map of
 /// keys/values that are in 'm2' but not in 'm1'.
-fn diff_maps<'a, K, V>(
+pub fn diff_maps<'a, K, V>(
```
NicolasDP: it seems this function could live in the standard library, up for a rustlang RFC ?

PR #357, opened 2018-12-25
Ho ho ho

NicolasDP:
# Merry Christmas

Now all new code must comply with default rust format.

i.e.:

```
rustup component add rustfmt-preview
cargo fmt
cargo fmt -- --check
```


mzabaluev:
Note that rustfmt has graduated out of the preview, so you should be using the `rustfmt` rustup component in stable.

NicolasDP:
thank @mzabaluev

PR #359, opened 2019-01-04
add chain-core crate

NicolasDP:
This new crate will defines the core objects that are used in the blockchain and the necessary properties.

### DIFF
--- a/chain-core/src/lib.rs
+++ b/chain-core/src/lib.rs
@@ -1,0 +1,12 @@
+#[macro_use]
+extern crate cfg_if;
+
+cfg_if! {
+    if #[cfg(test)] {
+        extern crate quickcheck;
+    } else if #[cfg(feature = "property-test-api")] {
+        extern crate quickcheck;
+    }
+}
+
+pub mod property;
--- a/chain-core/src/property.rs
+++ b/chain-core/src/property.rs
@@ -206,8 +206,8 @@ pub trait Serializable: Sized {
     fn deserialize<R: std::io::Read>(reader: R) -> Result<Self, Self::Error>;
 }
 
-#[cfg(test)]
-pub mod test {
+#[cfg(feature = "property-test-api")]
+pub mod testing {
     use super::*;
     use quickcheck::Arbitrary;
     use std::io::Cursor;
--- a/chain-core/src/property.rs
+++ b/chain-core/src/property.rs
@@ -1,0 +1,228 @@
+//! chain core properties
+//!
+//! define the different properties a _supported_ chain needs to
+//! implement to work in our models.
+//!
+//! # Block
+//!
+//! The Block is the atomic element that compose a chain. Or in other
+//! words the chain is composed of a succession of `Block`.
+//!
+//! the `Block` trait implements the necessary feature we expect of
+//! a `Block` in the chain. Having a function that requires the object
+//! to implement the Block traits means that we are expecting to have
+//! only access to:
+//!
+//! * the block and its parent's identifier (the block hash);
+//! * the block number, its position in the blockchain relative
+//!   to the beginning of the chain. We often call this number
+//!   the block Date.
+//!
+//! # HasTransaction and Transaction
+//!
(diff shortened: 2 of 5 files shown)


chain-core/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+#[macro_use]
+extern crate cfg_if;
+
+cfg_if! {
+    if #[cfg(test)] {
+        extern crate quickcheck;
+    } else if #[cfg(feature = "property-test-api")] {
```
mzabaluev: The feature is named `property-test-apis` in Cargo.toml. I think this name is better, though.

So you think it may be useful to provide quickcheck APIs to non-dev dependencies? This PR in current state does not use the feature to do it, but exports the test module in the `test` configuraition, which I think is not ideal (see below).

chain-core/src/property.rs
```diff
@@ -0,0 +1,228 @@
...
+
+/// accessor to transactions within a block
+///
+/// This trait is generic enough to show there is multiple types
+/// of transaction possibles:
+///
+/// * UTxO
+/// * certificate registrations
+/// * ...
+pub trait HasTransaction<T: Transaction> {
```
mzabaluev: This name is a bit odd. But then, we never came around to sufficiently generalizing this API, so maybe we can redesign this later.

chain-core/src/property.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// and the index of the output within the output array.
+    /// If the Output is not present it is possible that it does not exist or has
+    /// already been spent in another transaction.
+    fn input<'a>(&'a self, input: &T::Input) -> Result<&'a T::Output, Self::Error>;
+
+    /// create a new Update from the given transaction.
+    fn diff_transaction(&self, transaction: &T) -> Result<Self::Update, Self::Error>;
+
+    /// create a combined Update from the given transactions
+    fn diff<'a, I>(&self, transactions: I) -> Result<Self::Update, Self::Error>
```
mzabaluev: This could probably have a default implementation in terms of `diff_transaction` method.

chain-core/src/property.rs
```diff
@@ -0,0 +1,228 @@
...
+pub trait Serializable: Sized {
+    type Error: std::error::Error + From<std::io::Error>;
+
+    fn serialize<W: std::io::Write>(&self, writer: W) -> Result<(), Self::Error>;
+
+    fn deserialize<R: std::io::Read>(reader: R) -> Result<Self, Self::Error>;
+}
+
+#[cfg(test)]
+pub mod test {
```
mzabaluev: I would rather name the public quickcheck API module something different, and reserve the `test` module for unit tests, as per convention.

chain-core/src/property.rs
NicolasDP: you know what, isn't this type a bit redundant with `IntoIterator` ?

chain-core/src/property.rs
NicolasDP: indeed, the only constraint I see is that we will need to add something on the `Self::Update` type.

```rust
    type Update: std::ops::Add<Self::Update>;
```

chain-core/src/property.rs
mzabaluev: I thought that too, but it's probably not very ergonomic unless `IntoIterator` is implemented for the block object's reference.

NicolasDP: or even, with `AsRef<[T]> where T: Transaction`

chain-core/src/property.rs
mzabaluev: Ah yes, there would need to be a generic reduction op indeed.

chain-core/src/property.rs
mzabaluev: There may be an elegant generic way there that also clicks together with other methods consuming an `IntoIterator<Item = &Transaction>`. I should look into that, but as this is mostly a code move, it does not have to be done here.

PR #369, opened 2019-01-07
updated config.yml

amias-iohk:
added junit test output

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,7 @@ jobs:
           paths:
             - /usr/local/cargo/registry
             - /usr/local/cargo/git
+            - /usr/local/cargo/bin/cargo-junit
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -21,6 +22,7 @@ jobs:
           paths:
             - /usr/local/cargo/registry
             - /usr/local/cargo/git
+            - /usr/local/cargo/bin/cargo-junit
```
NicolasDP: Just reading the documentation of the `cargo-junit` crate: https://github.com/CWood1/cargo-junit#usage

and just like every cargo extension you need to use `cargo` to launch your extension once the extension installed.

i.e.:

```
cargo install cargo-junit
cargo junit --name test_output.xml
```

PR #371, opened 2019-01-07
Introduce basic version of the mockchain.

qnikst:


chain-impl-mockchain/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+          ]
+edition = "2018"
+
+[dependencies]
+serde = "^1.0"
+serde_derive = "^1.0"
+bincode = "^1.0.1"
+cryptoxide = "*"
+quickcheck = { version = "0.7", optional = true }
+chain-core = { path = "../chain-core", features=["property-test-api"]}
```
NicolasDP: this feature should only be enabled in the `dev-dependencies`

chain-impl-mockchain/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+extern crate serde;
+#[macro_use]
+extern crate serde_derive;
+
+pub mod mock;
```
NicolasDP: don't call this module `mock`. We already know that we are using the mockchain.

can you split the module actually ?

* error (with Result/Error defined in there);
* key (PublicKey, PrivateKey, Signature...);
* address (Address);
* transactions (TransactionId, Transaction, SignedTransaction, Witness, UtxoPointer, Output...);
* Block (Block, BlockId, SlockId, ...)

thanks

chain-impl-mockchain/Cargo.toml
qnikst: Done

chain-impl-mockchain/src/lib.rs
qnikst: Done

qnikst:
@NicolasDP please take another look.

PR #374, opened 2019-01-08
Introduce more tests for the mockchain.

qnikst:


qnikst:
@NicolasDP @mzabaluev please take a look, PR is passing now.

PR #376, opened 2019-01-08
Property traits for blockchain data entities

### LINKED ISSUE #375: Property traits to identify blockchain entity types
In code that is implemented in terms of the abstract blockchain, there is often a need to restrict generic implementations to the specific entities. Examles can be seen in #365: marshalling of a block ID into an opaque `BlockId` protocol message can technically work with any `Serialize` type, but is only meaningful for the type representing the block ID in the blockchain implementation.

Traits in mod `property` of `chain-core`, such as `Block` and `Transaction`, are already useful to identify some of the entities, though their main purpose is to provide API methods. For the other chain types, empty identification traits can be defined: `BlockId`, `BlockDate`, `TxId` etc. If these entities would need some abstract API of their own, their property traits would also be the natural place to add the methods.

mzabaluev:
Add empty property traits that are used to denote types representing
specific entities of the blockchain data model:

- `BlockId` implemented by block identifier type,
- `BlockDate` implemented by block date type,
- `TxId` implemented by transaction identifier type,
- `Header` implemented by block header type.

These continue the system of existing entity identification traits
such as `Block`, `Transaction`, and `Ledger`, except that the new traits
do not (yet) feature any methods.

Resolves #375

### DIFF
--- a/chain-core/src/property.rs
+++ b/chain-core/src/property.rs
@@ -35,6 +35,21 @@
 //! is selected to write a block in the chain.
 //!
 
+use std::hash::Hash;
+
+/// Trait identifying the block identifier type.
+pub trait BlockId: Eq + Hash {}
+
+/// Trait identifying the block date type.
+pub trait BlockDate: Eq + Ord {}
+
+/// Trait identifying the transaction identifier type.
+pub trait TransactionId: Eq + Hash {}
+
+/// Trait identifying the block header type.
+/// TODO: provide header in the data model.
+pub trait Header {}
+
 /// Block property
 ///
 /// a block is part of a chain of block called Blockchain.
@@ -51,13 +66,13 @@ pub trait Block: Serializable {
     ///
     /// In bitcoin this block is a SHA2 256bits. For Cardano's
     /// blockchain it is Blake2b 256bits.
-    type Id;
+    type Id: BlockId;
 
     /// the block date (also known as a block number) represents the
     /// absolute position of the block in the chain. This can be used
     /// for random access (if the storage algorithm allows it) or for
     /// identifying the position of a block in a given epoch or era.
-    type Date;
+    type Date: BlockDate;
 
     /// return the Block's identifier.
     fn id(&self) -> Self::Id;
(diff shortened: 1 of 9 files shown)


chain-core/src/property.rs
```diff
@@ -35,6 +35,21 @@
...
+use std::hash::Hash;
+
+/// Trait identifying the block identifier type.
+pub trait BlockId: Eq + Hash {}
+
+/// Trait identifying the block date type.
+pub trait BlockDate: Eq + Ord {}
+
+/// Trait identifying the transaction identifier type.
+pub trait TxId: Eq + Hash {}
```
NicolasDP: I would rather like to see `TransactionId` here. I know this was not the norm before but being more explicit to avoid confusion would help.

chain-core/src/property.rs
```diff
@@ -35,6 +35,21 @@
...
+
+/// Trait identifying the block header type.
+/// TODO: provide header in the data model.
+pub trait Header {}
```
NicolasDP: no we cannot provide this. the Header type is very specific to a given implementation of the blockchain. This is only needed for your implementation of the network. Can we keep it in there (if you want with a Todo reminding us that this may need to be generalised).

chain-impl-mockchain/src/block.rs
```diff
@@ -1,6 +1,6 @@
 //! Representation of the block in the mockchain.
-use crate::key::*;
-use crate::transaction::*;
+use crate::key::Hash;
+use crate::transaction::SignedTransaction;
```
NicolasDP: non necessary changes.

chain-core/src/property.rs
mzabaluev: This is needed for the `GetHeaders` gRPC method. Should we drop it altogether, or should it be optionally supported, depending perhaps on a dynamic property of `network_core::Node`?

chain-impl-mockchain/src/block.rs
mzabaluev: Yeah, I just jumped on the bad style. @qnikst, please avoid using glob imports except for prelude modules that are specifically designed for that.

qnikst: @mzabaluev ok, will do.

chain-core/src/property.rs
```diff
@@ -35,6 +35,21 @@
 //! is selected to write a block in the chain.
 //!
 
+use std::hash::Hash;
```
NicolasDP: Can this be move to the top of the file ?

mzabaluev: My IDE complained; I understand that module-level doc should go first, and it feels more logical to me too.

PR #378, opened 2019-01-09
Chain impl cardano

NicolasDP:


cardano/src/block/block.rs
```diff
@@ -200,6 +202,51 @@ impl fmt::Display for Block {
...
 
+impl core::property::Block for Block {
+    type Id = HeaderHash;
+    type Date = BlockDate;
+
+    fn id(&self) -> Self::Id {
+        self.get_header().compute_hash()
+    }
+
+    fn parent_id(&self) -> &Self::Id {
```
mzabaluev: NB: The by-reference return type looks a bit quirky here. It's inconsistent with the by-value return type of `Block::id` and leaks an assumption that the parent ID is an intrinsic value of every `Block` type. Making this method return by value should not be too much of a problem, as any conceivable `Id` type should either fit in a single cache line on most CPUs, or encapsulate a dynamically allocated buffer which is not copied with the value move.

PR #381, opened 2019-01-09
Move ChainStore from jormungandr to rust-cardano

edolstra:
Main changes:

* New chain_core properties: `ChainState` and `ChainStateDelta`, representing the abstract notion of a serializable diff between chain states.

* New crate `chain-storage` providing traits `BlockStore` and `ChainStateStore`, and an implementation `MemoryBlockStore` for in-memory chains.

* New crate `chain-storage-sqlite` that provides a `SQLiteBlockStore` implementation of `BlockStore`. Also provides two example programs that show creating a `SQLiteBlockStore` from an old `cardano-storage` chain and doing validation.

* Moved `cardano::ChainState` into a separate file. Also moved `diff_maps` into cardano::util since the `ChainStateStore` implementation of `cardano::ChainState` requires it.

* There now is a `BlockDate` trait (in `property.rs`) that requires `Eq`, `Ord` and `Clone` and requires the date to be serializable as a 64-bit integer. (This could be changed to `Serializable`, but a 64-bit integer is enough for now and more efficient in the SQLite backend since it doesn't require storing dates as blobs.)

* Also, a `BlockId` trait that imposes various requirements on IDs, e.g. they should be comparable. `BlockStore` needs this.


chain-storage-sqlite/examples/import.rs
```diff
@@ -0,0 +1,92 @@
...
+extern crate cardano;
+
+use chain_core::property::Block;
+use cardano_storage::{StorageConfig};
+use chain_storage::store::{BlockStore};
+use std::time::{Instant};
+use std::str::FromStr;
+
+fn main() {
+    let storage_config = StorageConfig::new(&"/home/eelco/.local/share/cardano-cli/blockchains/staging".into());
```
sjmackenzie: path issue

chain-storage-sqlite/examples/import.rs
```diff
@@ -0,0 +1,92 @@
+extern crate cardano_storage;
```
NicolasDP: It seems to me this could be moved to the test directory

cardano/src/block/block.rs
```diff
@@ -202,48 +202,46 @@ impl fmt::Display for Block {
     }
 }
 
-impl core::property::Block for Block {
+impl chain_core::property::Block for Block {
```
NicolasDP: why changing this ?

chain-core/src/property.rs
```diff
@@ -204,6 +222,58 @@ pub trait Serializable: Sized {
...
+
+    /// Update the chain state by applying a new 'block'. If an error
+    /// occurs, the chain state must be unchanged. If it succeeds,
+    /// then the chain length will be increased by 1 and
+    /// self.get_last_block_id() == block.id().
+    fn apply_block(&mut self, block: &Self::Block) -> Result<(), Self::Error>;
+
+    fn get_last_block_id(&self) -> <Self::Block as Block>::Id;
+
+    fn get_chain_length(&self) -> u64;
```
NicolasDP: usize in this case?

chain-storage-sqlite/examples/import.rs
```diff
@@ -0,0 +1,92 @@
...
+    let storage = cardano_storage::Storage::init(&storage_config).unwrap();
+
+    let genesis_hash = cardano::block::HeaderHash::from_str(&"c6a004d3d178f600cd8caa10abbebe1549bef878f0665aea2903472d5abf7323").unwrap();
+
+    let first_hash = cardano::block::HeaderHash::from_str(&"b365f1be6863b453f12b93e1810909b10c79a95ee44bf53414888513fe172c90").unwrap();
+
+    let tip_hash = cardano::block::HeaderHash::from_str(&"159e3cfe147dfbe302daf37b63dbce2e676cd88c1419f693ad9ebdc69cf4bc1c").unwrap();
+
+    let mut store = chain_storage_sqlite::SQLiteBlockStore::new(
+        genesis_hash, "/tmp/chain.sqlite");
```
amias-iohk: this path seems a bit inflexible and inherently temporary , shouldn't it use chain-storage variables that where defined in the cargo.toml ?

NicolasDP:
As you can see, in the `chain-core` there is a trait called `Ledger<T>`. The plan is to use this trait to manage the transaction verification and the transaction states. It focuses on a single element of the chain state : the block content, i.e. the transactions.

A new trait coming soon will be the one for the Update mechanism . The idea will be to validate votes, update proposals and to update the settings of the chain (in the case of the cardano blockchain).

We are doing so to have separation of concerns. But it is important to separate the properties as we need:

* independently unit testing the property's implementations;
* in Jormungandr the tasks running the different verifications and the different change of states are running on different threads as we are validating the received object at different levels (the transactions are validated in the Transaction Pool manager, the blocks are verified at the Block Query task, the Consensus is run by the Leadership task.

Now the ChainState you have provided seems to break this separation of concerns where the cardano impl seems to do every of the different properties.

* `Ledger` manages the content of the Block (i.e. the Transaction). It validates them and update its state;
* `LeaderSelection` manages the consensus algorithm;
* `ChainSettings` **will** manage the Update side of the blockchain (proposals, votes, ...);

Now here we are still missing the Block validation. And I think this might needs to be in a `Chain` trait to validate that every block are valid (checking the proofs, signatures...).

The UTxO state you need to maintain for the storage should fit in the `Ledger` trait.

edolstra:
Separating those properties for storage seems quite tricky to me since it can lead to confusion about whether a block is valid. E.g. a block could be valid with respect to the ledger state but not with respect to the update state. Also, maintaining three (or more) state stores is more complicated than maintaining one.

chain-core/src/property.rs
edolstra: No, usize is platform dependent which I don't think is desirable here.

cardano/src/block/block.rs
edolstra: I had to resolve the merge conflict with my own implementation of property::Block for Block, so this is what I picked. Using a consistent name for a crate seems a bit preferable.

chain-storage-sqlite/examples/import.rs
edolstra: No, unfortunately it can't be a test since there is no test data to apply it to.

chain-core/src/property.rs
```diff
@@ -35,13 +35,23 @@
...
 /// Trait identifying the block identifier type.
-pub trait BlockId: Eq + Hash {}
+pub trait BlockId: Eq + Ord + Clone + Debug + Hash + AsRef<[u8]> {
+    // FIXME: constant representing id length?
 
-/// Trait identifying the block date type.
-pub trait BlockDate: Eq + Ord {}
+    /// Construct a BlockId from a slice. FIXME: use TryFromSlice trait.
+    fn try_from_slice(slice: &[u8]) -> Option<Self>;
+}
```
NicolasDP: Why not use `BlockId : Serialize + Deserialize` ?

cardano/src/block/block.rs
```diff
@@ -231,32 +232,30 @@ impl core::property::Block for Block {
...
 }
 
-impl core::property::Deserialize for Block {
+impl chain_core::property::Deserialize for Block {
     type Error = cbor_event::Error;
 
     fn deserialize<R: std::io::BufRead>(reader: R) -> Result<Self, Self::Error> {
-        let mut deserializer = cbor_event::de::Deserializer::from(reader);
-        deserializer.deserialize::<Self>()
+        Deserialize::deserialize(&mut Deserializer::from(reader))
```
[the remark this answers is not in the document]
edolstra: This (and the other ones) are from my version of the Block trait implementation that I wrote before yours. I arbitrarily picked my version resolving the merge conflict.

chain-core/src/property.rs
```diff
@@ -35,13 +35,23 @@
...
+
+/// A trait representing block dates. Dates can be compared, ordered
+/// and serialized as integers.
+pub trait BlockDate: Eq + Ord + Clone {
+    fn serialize(&self) -> u64;
+    fn deserialize(n: u64) -> Self;
+}
```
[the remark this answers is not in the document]
edolstra: Hm, I thought I added a comment about that but apparently it got lost somewhere. The reason is that serializing as arbitrary-length byte vectors requires storing as a blob in in the sqlite backend, which is more expensive.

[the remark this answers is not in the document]
NicolasDP: Ok make sense. Bit manipulation has a special place in my heart. How about having `From<u64>` and `Into<u64>`  as requirements of the trait instead ? The naming of the function is clashing with stream serialisation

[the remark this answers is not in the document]
NicolasDP: yes, lets get rid of it then

chain-storage/src/store.rs
```diff
@@ -198,21 +225,29 @@ pub trait BlockStore<B>: std::marker::Sized where B: Block {
...
             }
         }
     }
 }
 
-pub struct BlockIterator<'store, B, S> where B: Block, S: BlockStore<B> + 'store {
+pub struct BlockIterator<'store, B, S>
+where
+    B: Block,
+    S: BlockStore<B> + 'store,
```
mzabaluev: Why must `BlockStore` have the block type as a parameter and not as an associated type? I think this percolates into excessive generic type parameters in other places, such as here.

chain-core/src/property.rs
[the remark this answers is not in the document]
mzabaluev: As we deal with an abstract chain, I'm not sure all possible block dates can have an isomorphic map to `u64`.

This shows that we may have a need for two different serializations: one for the network protocol and the other for storage. In fact, for gRPC I think to use the string representation of the blockdate as implemented by traits `FromStr` and `ToString`.

chain-storage/src/store.rs
edolstra: Thanks, changed it into an associated type.

PR #384, opened 2019-01-09
Introduces SignedBlock for the mockchain.

qnikst:


chain-impl-mockchain/src/block.rs
```diff
@@ -62,6 +64,98 @@ impl property::HasTransaction<SignedTransaction> for Block {
...
+
+/// `Block` is an element of the blockchain it contains multiple
+/// transaction and a reference to the parent block. Alongside
+/// with the position of that block in the chain.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub struct SignedBlock {
+    /// Internal block.
+    block: Block,
+    /// List of cryptographic signatures that verifies the block.
+    signatures: Vec<(Address, Signature)>,
```
NicolasDP: there is only one signature in our case, no need for address.

we just need:

```rust
struct SignedBlock {
    block: Block,
    signatures: Signature,
    public_key: PublicKey,
}
```

chain-impl-mockchain/src/block.rs
```diff
@@ -62,6 +64,98 @@ impl property::HasTransaction<SignedTransaction> for Block {
...
+}
+
+impl SignedBlock {
+    /// Create a new signed block.
+    pub fn new(block: Block) -> Self {
+        SignedBlock {
+            block: block,
+            signatures: vec![],
+        }
```
NicolasDP: no, so what you need to do here (once the type is right) is:

```rust
impl SignedBlock {
    pub fn new(key: &PrivateKey, block: Block) -> Self {
        let id = block.id();
        let signature = key.sign(id.as_ref());
        Block { block, signature, public_key: key.public() }
    }
```

qnikst:
@NicolasDP please take another look, I've addressed your comments.

chain-impl-mockchain/src/block.rs
```diff
@@ -69,6 +70,80 @@ impl property::HasTransaction<SignedTransaction> for Block {
...
+    type Date = <Block as property::Block>::Date;
+    type Header = <Block as property::Block>::Header;
+
+    /// Identifier of the block, currently the hash of the
+    /// serialized transaction.
+    fn id(&self) -> Self::Id {
+        self.block.id()
+    }
+
+
```
NicolasDP: rustfmt is going to fail on this line. please remove.

qnikst: Fixed

PR #387, opened 2019-01-09
disabled email notifications from appveyor

amias-iohk:
changed the syntax for disabling emails from appveyor , looks like our last one didn't seem to work properly

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -90,9 +90,10 @@ branches:
 
 notifications:
   - provider: Email
-    to:nobody@iohk.io
+    to:
+      - nobody@iohk.io
     on_build_success: false
-    on_build_failre: false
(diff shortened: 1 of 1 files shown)


amias-iohk:
this is working and ready to merge

amias-iohk:
the notification was disabled some 6 months ago , it seems this wasn't done correctly and it broke the config file which stopped the builds running. My change fixes the config file which gets the builds to work again, up to the point where another error breaks them.

email notifications for these builds aren't really needed because we see them in the PR checks , it can create confusion because they arrive a while after the fact.

NicolasDP:
I don't see how it is not working prior this change. Other PRs have no problem with the appveyor builds (except the current rust nightly job breakdown...)

PR #390, opened 2019-01-10
mockchain: add more tests.

qnikst:


qnikst:
PR passes everything except for nightly so is ready for review.

PR #391, opened 2019-01-11
Crate network-core

### LINKED ISSUE #389: Add crate network-core to provide abstract network APIs
The network protocol abstractions that have been started in a `network_core` module in `network-grpc` need to be migrated into a dedicated library crate that will be shared with the NTT protocol implementation (#388).

mzabaluev:
Moved the abstract network interface definitions, incubated in
network-grpc, to a new dedicated crate network-core.

The interfaces rely on enum types generated from protobuf descriptions.
For this, the protobuf declarations of enums have been split out into
a dedicated package codes.proto which is used by the build script of
network-core to generate the types and make them available as
module codes.

Resolves #389


network-core/src/server.rs
```diff
@@ -21,13 +22,13 @@ impl fmt::Display for Error {
...
 }
 
 pub struct RecordTransactionResponse<Id> {
     // TODO: define
     _id: Id,
-    _result: gen::record_transaction_response::Result,
+    _result: codes::TransactionAcceptance,
 }
 
 pub trait Node {
```
NicolasDP: the trait is a bit off the generalisation. The Error type is a bit dependant on what the protocol implements. In cardano legacy's network they defined Error type that are defined on the command basis. This is something the implementation of the network should provide actually.

Also why not separating a bit the concerns too:

1. we can have a trait for Connection which is the one responsible to provide the connection/shutdown/reconnection/timeout logic;
2. we could have the Transaction trait to require the _propose transaction method_ and the _send transaction method_.
3. ...

This can allows us to, again, separate concerns but also to extend BlockStream to leverage the random access memory properties etc...

mzabaluev: Thanks for the insight. The abstract connection traits will bring back a notion of peers, which I lost when moving the code from jormungandr, and definitely concern separation will be useful. I can rework the traits and add more commits to this PR, or feel free to merge this one and I will create another.

mzabaluev: One reason I have made `Error` a single concrete type is that this simplified conversions to gRPC status messages. As you mention, this is overly simplistic. I think we can model this with different error types for each concern-separated service trait, but I'd like to avoid abstract associated error types because that would bring a need to supply M:N conversions between the actual types and protocol messages.

mzabaluev: > The abstract connection traits will bring back a notion of peers

On a later thought, peers are mainly important in the client-side operation of the node, and the server should not have per-peer state besides perhaps what is needed for traffic/rate limiting.

mzabaluev: I get corrected in team channel discussion: we need a notion of connection on the server side as well, because the protocol connections can be bidirectional and if they are, we don't want to open a duplicate connection to talk as a client to the same peer node. Subscription management can involve connections as well, though subscriptions can be modelled in simply as long-running sinks with a specific error case for when the client subscription has been torn down.

mzabaluev: I have split the `Node` server trait in f5d08b77a6bf4ba27a9daf0d988d96951fc2b531 to separate service traits implementing different faces of the node protocol.

PR #392, opened 2019-01-11
More lightweight access to block header properties

### LINKED ISSUE #385: cardano: Block::get_header too heavy for many uses
The method `Block::get_header` in the `cardano` chain data model retrieves the header by value, cloning all header data from the variants of `Block` into corresponding variants of `BlockHeader`. A brief audit of uses of the method shows that many of them use the return value temporarily to retrieve header properties, then drop the copied data.

The chain API can be redesigned so that `Block` provides direct property accessor methods with more lightweight return values. The implementations of these methods can destructure the `Block` variants and their variant-specific header structures to get at the value to clone and return.

`get_header` could then be deprecated (also for its un-Rustic naming convention), and a new method can be provided instead, with name and perhaps signature tailored to hint the user that extraction of the owned `BlockHeader` value is expensive and should not be taken lightly.

mzabaluev:
Fixes #385

### DIFF
--- a/cardano/src/block/block.rs
+++ b/cardano/src/block/block.rs
@@ -58,7 +58,7 @@ impl RawBlock {
     pub fn to_header(&self) -> cbor_event::Result<RawBlockHeader> {
         // TODO optimise if possible with the CBOR structure by skipping some prefix and some suffix ...
         let blk = self.decode()?;
-        Ok(blk.get_header().to_raw())
+        Ok(blk.header().to_raw())
     }
 }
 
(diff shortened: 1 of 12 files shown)


cardano/src/block/block.rs
```diff
@@ -295,6 +368,19 @@ impl cbor_event::de::Deserialize for Block {
...
+        serializer: &'se mut Serializer<W>,
+    ) -> cbor_event::Result<&'se mut Serializer<W>> {
+        let serializer = serializer.write_array(cbor_event::Len::Len(2))?;
+        match self {
+            BlockHeaderView::Boundary(hdr) => serializer.write_unsigned_integer(0)?.serialize(hdr),
+            BlockHeaderView::Normal(hdr) => serializer.write_unsigned_integer(1)?.serialize(hdr),
+        }
+    }
+}
+
```
NicolasDP: can you add a test for the BlockHeaderView serialize actually deserialize to the expected BlockHeader ?

mzabaluev: Good idea, will do. Should I use property test grinding?

PR #398, opened 2019-01-17
WIP: Move network-transport implemntation to the network-core

qnikst:
PR is done for an early review and has lots of work to do:

 * [ ] implement streaming replies
 * [ ] generalize messages, remove dependency on cardano
 * [ ] implement logging
 * [ ] implement subscription mechanism
 * [ ] verify that network can be injected into the loop in application
 * [ ] remove redundant constraints on the Node and it's associated types.

### DIFF
--- a/network-ntt/src/lib.rs
+++ b/network-ntt/src/lib.rs
@@ -1,0 +1,9 @@
+//! all the network related actions and processes
+//!
+//! This module only provides and handle the different connections
+//! and act as message passing between the other modules (blockchain,
+//! transactions...);
+
+extern crate protocol_tokio as protocol;
+
+pub mod server;
--- a/network-ntt/src/server.rs
+++ b/network-ntt/src/server.rs
@@ -1,0 +1,412 @@
+//! Network transport node, implementes an interface to the
+//! chain node, based on the network-transport protocol (NT later).
+//! NT is a simplified protocol that mimic CCI, a protocol
+//! that support differrent quality networks and provides a high
+//! level interface for the mutiplexing uni-directional connections
+//! between the nodes.
+//!
+//! For better understanding of the network transport concept
+//! you may check [document]().
+//!
+//! Node may act both as a server and a client. Server listens
+//! on the specified port and client connects to the remote service.
+//! Once node node establised a connection to the other one, the other
+//! once can build an opposite connection and nodes can talk to
+//! each other.
+//!
+//!```text
+//!                                      +---- server: sock1
+//!                                      |      |
+//!                                      |      |-- connection1
+//!                                      |      |-- connection2
+//!   +--------------+     +----------+  |
+//!   | network-core | ----| NT state |--+---- server: sock2
+//!   +--------------+     +----------+  |      |
+//!                                      |      |-- connection3
+//!                                      |
+//!                                      +---- client: remote-sock1
+//! ```
+//!
+//! As NT state is shared all the services could provide a uniform
+//! access to the chain state. And one could build very flexible
+//! topology.
+use core::marker::PhantomData;
+use futures::{future, prelude::*, stream::Stream, sync::mpsc};
+use network_core::server::{self, block::BlockService, transaction::TransactionService};
+use protocol::{network_transport::LightWeightConnectionId, Inbound, Message};
+use std::net::SocketAddr;
+use tokio::net::{TcpListener, TcpStream};
+
+// These types are loaded from cardano package,
+// because tokio-protocol is not generalized yet.
+// This should be fixed and then these types should go.
+use cardano::{
+    block::{Block, BlockHeader, HeaderHash},
+    tx::TxAux,
+};
+
+/// Internal structure of network transport node.
+#[derive(Clone)]
+pub struct Server<N> {
+    node: N,
+}
+
+/// Sets up a listening TCP socket bound to the given address.
+/// If successful, returns an asynchronous stream of `TcpStream` socket.
+pub fn listen(
+    sockaddr: SocketAddr,
+) -> Result<impl Stream<Item = TcpStream, Error = tokio::io::Error>, tokio::io::Error> {
+    let listener = TcpListener::bind(&sockaddr)?;
+    let stream = listener.incoming();
+    Ok(stream)
+}
+
+//pub struct Connection<N,F>(F,Server<N>)
+pub struct Connection<F>(F)
+where
+    F: future::Future<Item = (), Error = ()>;
+
+//impl<N,F> Future for Connection<N,F>
+impl<F> Future for Connection<F>
+where
+    F: future::Future<Item = (), Error = ()>,
+{
+    type Item = ();
+    type Error = ();
+
+    fn poll(&mut self) -> Poll<(), ()> {
+        self.0.poll()
+    }
+}
+
+trait SendMsg {
+    fn mk_ok_msg(id: LightWeightConnectionId, msg: Self) -> Message;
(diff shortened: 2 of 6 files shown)


network-ntt/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+//!
+//! This module only provides and handle the different connections
+//! and act as message passing between the other modules (blockchain,
+//! transactions...);
+
+//extern crate cardano;
+//#[macro_use]
+//extern crate futures;
+//extern crate tokio;
+extern crate protocol_tokio as tokio_protocol;
```
mzabaluev: Just use `protocol` for brevity?

mzabaluev: Also, `extern crate` is unnecessary in Rust 2018.

network-ntt/src/reply_handle.rs
```diff
@@ -0,0 +1,120 @@
+//! This module provides an abstraction to a simple connection
+//! reply mechanism. NT provides a very cheap lightweight connection,
```
mzabaluev: I intended for `network-core` future and stream implementations to hide the details of serving the request. The `Server` implementation should be able to resolve requests with its embedded `Node` object.

network-ntt/src/server.rs
```diff
@@ -0,0 +1,313 @@
...
+//!                                     |      |
+//!                                     |      |-- connection1
+//!                                     |      |-- connection2
+//!  +--------------+     +----------+  |
+//!  | network-core | ----| NT state |--+---- server: sock2
+//!  +--------------+     +----------+  |      |
+//!                                     |      |-- connection3
+//!                                     |
+//!                                     +---- client: remote-sock1
+//! ```
```
mzabaluev: Cool. Four-space indent should works too.

network-ntt/src/server.rs
```diff
@@ -0,0 +1,313 @@
...
+//! As NT state is shared all the services could provide a uniform
+//! access to the chain state. And one could build very flexible
+//! topology.
+//use crate::reply_handle::ReplyHandle;
+//use chain_core::property::{Deserialize, Serialize};
+use futures::prelude::*;
+use futures::{future, stream::{self,Stream}, sync::mpsc};
+use network_core::server;
+use network_core::server::block::BlockService;
+use network_core::server::transaction::TransactionService;
```
mzabaluev: I tend to group imports by namespace with brackets, but I have no strong opinion about this.

network-ntt/src/server.rs
```diff
@@ -0,0 +1,313 @@
...
+#[derive(Clone)]
+pub struct Server<N> {
+    node: N,
+}
+
+/// Run NTT server.
+pub fn run<N: 'static>(config: Config, node: N)
+where
+    N: std::marker::Sync + std::marker::Send + server::Node + Clone + Copy,
+    <<N as network_core::server::Node>::BlockService as network_core::server::block::BlockService>::TipFuture: std::marker::Send,
```
mzabaluev: You can shorten the `network_core::server` paths to just `server`. Maybe it would be good to `use network_core::server as core_server` at the top for less ambiguity.

network-ntt/src/server.rs
```diff
@@ -0,0 +1,313 @@
...
+
+        let listener =
+            stream::iter_ok(config.listen)
+                .for_each(move |sockaddr| {
+                    let server = Server {
+                          node: node.clone(),
+                    };
+                    run_listen_socket(sockaddr, server) 
+                });
+        let connections =
```
mzabaluev: There should be a `Connection` object to encapsulate peer connection state, like in `network-grpc`

network-ntt/src/server.rs
```diff
@@ -0,0 +1,313 @@
...
+// These types are loaded from cardano package,
+// because tokio-protocol is not generalized yet.
+// This should be fixed and then these types should go.
+use cardano::{
+    block::{HeaderHash},
+    tx::{TxAux},
+};
+
+/// Network transport server configuration.
+pub struct Config {
```
mzabaluev: I did not bring a `Config` into the `network-grpc` library. Individual calls to `server::listen` and a future `server::connect` can be used to instantiate connections. I think we could live it up to the application how to manage collections of peers, unless we have a generally perfect idea for everyone.

network-ntt/src/server.rs
```diff
@@ -0,0 +1,313 @@
...
+    <<N as network_core::server::Node>::BlockService as network_core::server::block::BlockService>::TipFuture: std::marker::Send,
+    <<N as network_core::server::Node>::BlockService as network_core::server::block::BlockService>::GetHeadersFuture: std::marker::Send,
+    <<N as network_core::server::Node>::BlockService as network_core::server::block::BlockService>::StreamBlocksToTipFuture: std::marker::Send,
+    <<N as network_core::server::Node>::TransactionService as network_core::server::transaction::TransactionService>::ProposeTransactionsFuture: std::marker::Send,
+{
+    let server = TcpListener::bind(&sockaddr)
+        .expect("Can't bind socket")
+        .incoming()
+        .map_err(move |_err| ())
+        .for_each(move |stream| {
```
mzabaluev: I recommend separating out the listener stream creation to a helper function,
like done in `network-grpc`. The idea is for the application to be able to examine incoming connections as bare socket objects and decide if it wants to create protocol connections around them.

I will add another PR where the transport connection type is generalized in network-grpc, allowing TCP and Unix sockets. This is where such design also comes helpful.

network-ntt/src/server.rs
```diff
@@ -0,0 +1,313 @@
...
+}
+
+/// Internal structure of network transport node.
+#[derive(Clone)]
+pub struct Server<N> {
+    node: N,
+}
+
+/// Run NTT server.
+pub fn run<N: 'static>(config: Config, node: N)
```
mzabaluev: This looks too monolithic to put it into a library.
The server application should be able to create or accept individual peer connections, and drive their futures in its preferred way. We will have connection task futures from both network-grpc and network-ntt at runtime, and we should be able to run them on the same thread pool for better resource utilization (if we ever manage to make them `Send`).

network-ntt/src/server.rs
```diff
@@ -0,0 +1,313 @@
...
+                    tokio::spawn(run_connection(node, connection))
+                })
+        });
+    tokio::spawn(server)
+}
+
+/// Connect to another client.
+/// `run_connection` is spawned on the single heavyweight
+/// connection.
+pub fn run_connect_socket<N: 'static>(
```
mzabaluev: I would put this away for another PR, when we design some sort of client API abstraction.

network-ntt/src/lib.rs
qnikst: `protocol` clashes with `protocol` crate.

network-ntt/src/server.rs
qnikst: I'd keep it here due to the nature of the n-t, the client in network transport implementation will be some other thing.

network-ntt/src/lib.rs
mzabaluev: That one is not imported, though? I don't think swapping words in the name brings more clarity to the code.

qnikst: Yeah that is fixed in the latest versions

network-ntt/src/server.rs
qnikst: Done

mzabaluev:
I think we should close this as the work has been continued in #401. Move the TODO checklist there if it's still needed.

PR #400, opened 2019-01-21
Update the setting Trait

NicolasDP:
add tip to the setting trait.

### DIFF
--- a/chain-core/Cargo.toml
+++ b/chain-core/Cargo.toml
@@ -11,10 +11,10 @@ edition = "2018"
 
 [dependencies]
 cfg-if = "0.1"
-quickcheck = { version = "0.7", optional = true }
+quickcheck = { version = "0.8", optional = true }
 
 [dev-dependencies]
-quickcheck = "0.7"
+quickcheck = "0.8"
 
 [features]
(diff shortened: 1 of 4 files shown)


chain-core/src/property.rs
```diff
@@ -257,6 +257,8 @@ pub trait Settings {
 
     /// apply the Update to the Settings
     fn apply(&mut self, update: Self::Update) -> Result<(), Self::Error>;
+
+    fn tip(&self) -> &<Self::Block as Block>::Id;
```
mzabaluev: No need to commit to keeping a reference to tip inside the type. Consider returning by value.

chain-core/Cargo.toml
```diff
@@ -11,10 +11,10 @@ edition = "2018"
 
 [dependencies]
 cfg-if = "0.1"
-quickcheck = { version = "0.7", optional = true }
+quickcheck = { version = "0.8", optional = true }
 
 [dev-dependencies]
-quickcheck = "0.7"
+quickcheck = "0.8"
```
mzabaluev: This seems unrelated, but refreshing dependencies is always welcome ;)

PR #401, opened 2019-01-22
Generalize network-ntt

qnikst:
This PR generalized and implements network transport node.

### DIFF
--- a/protocol-tokio/src/protocol/codec/message.rs
+++ b/protocol-tokio/src/protocol/codec/message.rs
@@ -373,10 +373,11 @@ impl<T: de::Deserialize + property::BlockId> de::Deserialize for GetBlocks<T> {
 #[cfg(test)]
 mod test {
     use super::*;
-    use cbor_event::de::Deserializer;
-    use std::io::Cursor;
+    use chain_core::property;
 
-    impl<T: ::quickcheck::Arbitrary> ::quickcheck::Arbitrary for GetBlockHeaders<T> {
+    impl<T: ::quickcheck::Arbitrary + property::BlockId> ::quickcheck::Arbitrary
+        for GetBlockHeaders<T>
+    {
         fn arbitrary<G: ::quickcheck::Gen>(g: &mut G) -> Self {
             let from = ::quickcheck::Arbitrary::arbitrary(g);
             let to = ::quickcheck::Arbitrary::arbitrary(g);
@@ -384,28 +385,11 @@ mod test {
         }
     }
 
-    impl<T: ::quickcheck::Arbitrary> ::quickcheck::Arbitrary for GetBlocks<T> {
+    impl<T: ::quickcheck::Arbitrary + property::BlockId> ::quickcheck::Arbitrary for GetBlocks<T> {
         fn arbitrary<G: ::quickcheck::Gen>(g: &mut G) -> Self {
             let from = ::quickcheck::Arbitrary::arbitrary(g);
             let to = ::quickcheck::Arbitrary::arbitrary(g);
             GetBlocks { from: from, to: to }
-        }
-    }
-    quickcheck! {
-        fn get_block_headers_encode_decode(command: GetBlockHeaders<u8>) -> bool {
-            let encoded = cbor!(command).unwrap();
-            let mut de = Deserializer::from(Cursor::new(&encoded));
-            let decoded : GetBlockHeaders<u8> = de.deserialize_complete().unwrap();
-
-            decoded == command
-        }
-
-        fn get_blocks_encode_decode(command: GetBlocks<u8>) -> bool {
-            let encoded = cbor!(command).unwrap();
-            let mut de = Deserializer::from(Cursor::new(&encoded));
-            let decoded : GetBlocks<u8> = de.deserialize_complete().unwrap();
-
-            decoded == command
         }
     }
 }


protocol-tokio/src/lib.rs
```diff
@@ -1,14 +1,15 @@
...
 extern crate bytes;
-extern crate cardano;
 extern crate tokio_codec;
 extern crate tokio_io;
 #[macro_use]
 extern crate futures;
 #[macro_use]
 extern crate log;
 
+extern crate network_core;
```
mzabaluev: I hope this is temporary and we will eventually de-spaghettify the crates by merging protocol-tokio with network-ntt.

protocol-tokio/src/protocol/codec/message.rs
```diff
@@ -4,10 +4,10 @@ use std::{
...
 };
 
-use cardano::{
-    block::{self, HeaderHash},
-    tx::TxAux,
-};
+//use cardano::{
+//    block::{self, HeaderHash},
+//    tx::TxAux,
+//};
```
mzabaluev: Just remove this.

protocol-tokio/src/protocol/codec/message.rs
```diff
@@ -88,17 +89,113 @@ pub enum Message {
...
+
+    UserMessage(M),
+}
 
-    GetBlockHeaders(nt::LightWeightConnectionId, GetBlockHeaders),
-    BlockHeaders(nt::LightWeightConnectionId, Response<BlockHeaders, String>),
-    GetBlocks(nt::LightWeightConnectionId, GetBlocks),
-    Block(nt::LightWeightConnectionId, Response<Block, String>),
-    SendTransaction(nt::LightWeightConnectionId, TxAux),
+pub enum ChainMessage<B, T>
```
mzabaluev: A comment explaining the purpose of `ChainMessage` would be nice.

protocol-tokio/src/protocol/codec/message.rs
```diff
@@ -88,17 +89,113 @@ pub enum Message {
...
+
+pub trait AsNtEvent
+where
+    Self: std::marker::Sized,
+{
+    fn to_nt(self) -> nt::Event;
+    fn try_from_nt(lwcid: nt::LightWeightConnectionId, bytes: &Bytes) -> Option<Self>;
+}
+
+pub struct Any();
```
mzabaluev: What is this for?

protocol-tokio/src/protocol/codec/message.rs
```diff
@@ -88,17 +89,113 @@ pub enum Message {
...
+            MessageType::MsgBlock => Some(ChainMessage::Block(
+                lwcid,
+                cbor.deserialize_complete().unwrap(),
+            )),
+            MessageType::MsgSubscribe1 => {
+                let v: u64 = cbor.deserialize_complete().unwrap();
+                let keep_alive = v == 43;
+                Some(ChainMessage::Subscribe(lwcid, keep_alive))
+            }
+            _ => unimplemented!(),
```
mzabaluev: This silently creates a crash on any accidentally unimplemented cases.
I'd rather add `unimplemented!()` cases on every variant in an exhaustive match.

protocol-tokio/src/protocol/codec/message.rs
```diff
@@ -88,17 +89,113 @@ pub enum Message {
...
+    GetBlocks(nt::LightWeightConnectionId, GetBlocks<B::BlockId>),
+    Block(nt::LightWeightConnectionId, Response<B::Block, String>),
+    SendTransaction(nt::LightWeightConnectionId, T::TransactionId),
     TransactionReceived(nt::LightWeightConnectionId, Response<bool, String>),
     Subscribe(nt::LightWeightConnectionId, KeepAlive),
-    Bytes(nt::LightWeightConnectionId, Bytes),
 }
-impl Message {
+
+pub trait AsNtEvent
```
mzabaluev: The naming of both `AsNtEvent` and `to_nt` is confusing with regard to the established naming conventions of conversion traits. `to_nt` should be `into_nt`,
and the trait should be not named with `As*` (implying a by-reference conversion).

Even more Rustic would be to turn the `to_nt` half into `impl From<Message> for nt::Event` and leave `try_from_nt` in a `TryFrom`-like trait (but only because the standard `TryFrom` is unstable).

protocol-tokio/src/protocol/codec/message.rs
```diff
@@ -88,17 +89,113 @@ pub enum Message {
     ProbeSocketAck(nt::LightWeightConnectionId),
     CreateNodeId(nt::LightWeightConnectionId, NodeId),
     AckNodeId(nt::LightWeightConnectionId, NodeId),
+    Bytes(nt::LightWeightConnectionId, Bytes),
+
+    UserMessage(M),
```
mzabaluev: This extension variant complicates the code quite a lot. I wonder if it's really needed.

protocol-tokio/src/protocol/inbound_stream.rs
```diff
@@ -111,31 +154,47 @@ impl<T> InboundStream<T> {
...
-            Message::TransactionReceived(lwcid, r) => {
-                self.forward_message(lwcid, Inbound::TransactionReceived, r)
-            }
-            Message::GetBlocks(lwcid, gb) => self.forward_message(lwcid, Inbound::GetBlocks, gb),
-            Message::Block(lwcid, b) => self.forward_message(lwcid, Inbound::Block, b),
-            Message::Subscribe(lwcid, keep_alive) => {
-                self.forward_message(lwcid, Inbound::Subscribe, keep_alive)
-            }
             Message::Bytes(lwcid, bytes) => self.forward_message(lwcid, Inbound::Data, bytes),
+            Message::UserMessage(msg) => {
```
mzabaluev: Does this effectively extend the wire protocol, on the network as well?

protocol-tokio/src/protocol/codec/message.rs
qnikst: I'll reply here (as that reply goes to the other comments as well). The idea of the network protocol is that is has some `technical/control` messages and user messages go over the wire. The idea of the change was to keep 2 types one for the general in `protocol` library and one user specific. In our case `Message` itself is general and `ChainMessage` is application specific. The alternative is that user will always implement his `Message` that will keep control messages around. There are some conventions how to implement such a message so it can be put on the wire flat (i.e. not reuse the prefixes of the control one).
So this change does not extend network layout, but it allows to write the protocol in a generic way.

protocol-tokio/src/protocol/codec/message.rs
qnikst: This is a hack, any better idea are welcome, the problem is that for the type inference rustc should know the type of the application specific payload, but `accepting` and `connecting` traits do not care about it, as they do not exchange the application specific payload. `Any` introduces a special type that we can use there so those structures work for any user payload.

network-core/src/server/block.rs
```diff
@@ -61,6 +61,21 @@ pub trait BlockService {
...
+        &mut self,
+        from: &[Self::BlockId],
+        to: &Self::BlockId,
+    ) -> Self::GetHeadersFuture;
+    // Stream blocks to the provided tip.
+    fn stream_blocks_to(
+        &mut self,
+        from: &[Self::BlockId],
+        to: &Self::BlockId,
+    ) -> Self::StreamBlocksToTipFuture;
```
mzabaluev: Good reuse. I'd then rename the type to `StreamBlocksFuture` for consistency.
(I also have a bikeshedding desire to rename these methods to `pull_blocks`, but this can be done later)

network-ntt/src/server.rs
```diff
@@ -0,0 +1,288 @@
...
+                        .map_err(|err| err.to_string())
+                        .and_then(move |headers| {
+                            Stream::collect(headers)
+                                .map_err(|err| err.to_string())
+                                .and_then(move |hdrs| {
+                                    let msg = Message::BlockHeaders(
+                                        lwcid,
+                                        Response::Ok(BlockHeaders(hdrs)),
+                                    );
+                                    sink1.unbounded_send(msg).map_err(|err| err.to_string())
```
mzabaluev: The errors are handled here...

network-ntt/src/server.rs
```diff
@@ -0,0 +1,288 @@
...
+                                })
+                        })
+                        .map_err(move |msg| {
+                            let msg = Message::BlockHeaders(lwcid, Response::Err(msg));
+                            sink2.unbounded_send(msg).unwrap_or_else(|_| ());
```
mzabaluev: ... but ignored here and in other `unbounded_send` call sites? I'd rather have an `.unwrap()` everywhere as these unbounded MPSC channel operations shouldn't fail in normal operation.

network-ntt/src/server.rs
```diff
@@ -0,0 +1,288 @@
...
+                }
+                Inbound::SendTransaction(_lwcid, tx) => {
+                    future::Either::B(future::Either::B(future::Either::B(
+                        server
+                            .node
+                            .transaction_service()
+                            .unwrap()
+                            .propose_transactions(&vec![tx])
+                            .and_then(|_| Ok(()))
+                            .or_else(|_| Ok(())),
```
mzabaluev: Um... I think there should be a FIXME comment to add proper error handling later, at least.

protocol-tokio/Cargo.toml
```diff
@@ -10,10 +10,9 @@ description = """
 Cardano full node network protocol using rust's futures for asynchronous
 network connection support.
 """
-keywords = [ "Cardano", "Protocol" ]
```
mzabaluev: "blockchain", the hottest keyword in 2019! :)
I think the keywords are canonically lowercase, but it may be unimportant.

protocol-tokio/src/protocol/accepting.rs
```diff
@@ -9,39 +9,48 @@ use tokio_io::{AsyncRead, AsyncWrite};
 use cbor_event::de::Deserializer;
 use std::{self, io::Cursor, vec};
 
-use super::{nt, Connection, Handshake, Message, NodeId};
+use super::{nt, raw_msg_to_nt, Connection, Handshake, Message, NodeId};
 
-enum AcceptingState<T> {
+enum AcceptingState<T, Header, BlockId, Block, TransactionId> {
```
mzabaluev: OK... I'm afraid I'll have to do something like this in network-grpc, but I'm trying to resolve it in terms of chain-core property traits.

protocol-tokio/src/protocol/codec/message.rs
```diff
@@ -192,7 +201,14 @@ impl Message {
     }
 }
 
-fn decode_node_ack_or_syn(lwcid: nt::LightWeightConnectionId, bytes: &Bytes) -> Option<Message> {
+pub fn raw_msg_to_nt(msg: Message<u8, u8, u8, u8>) -> nt::Event {
```
mzabaluev: Is this free function more useful than just calling the method?

protocol-tokio/src/protocol/mod.rs
```diff
@@ -65,16 +66,26 @@ impl ConnectionState {
...
 /// Once established call `split` to get the inbound stream
 /// and the outbound sink and starts processing queries
-pub struct Connection<T> {
+pub struct Connection<T, Header, BlockId, Block, TransactionId> {
     connection: nt::Connection<T>,
     state: Arc<Mutex<ConnectionState>>,
+    header: PhantomData<Header>,
+    block_id: PhantomData<BlockId>,
+    block: PhantomData<Block>,
+    transaction_id: PhantomData<TransactionId>,
```
mzabaluev: This is annoying. Maybe we should add a trait to network-core grouping all these types as associated, to have just one parameter here and in the protocol message types above.

protocol-tokio/src/protocol/codec/message.rs
```diff
@@ -240,11 +252,11 @@ impl<T: de::Deserialize, E: de::Deserialize> de::Deserialize for Response<T, E>
...
 #[derive(Debug, PartialEq, Eq, Clone)]
-pub struct GetBlockHeaders {
-    pub from: Vec<HeaderHash>,
-    pub to: Option<HeaderHash>,
+pub struct GetBlockHeaders<T> {
+    pub from: Vec<T>,
+    pub to: Option<T>,
 }
-impl GetBlockHeaders {
+impl<T> GetBlockHeaders<T> {
```
mzabaluev: It might make sense to restrict the type parameter to the chain-core classifier trait `property::BlockId`, to prevent accidental parameterization of `GetBlockHeader` with unintended types.

protocol-tokio/src/protocol/codec/message.rs
```diff
@@ -301,37 +314,42 @@ impl de::Deserialize for GetBlockHeaders {
 }
 
 #[derive(Clone, Debug)]
-pub struct BlockHeaders(pub Vec<block::BlockHeader>);
-impl From<Vec<block::BlockHeader>> for BlockHeaders {
-    fn from(v: Vec<block::BlockHeader>) -> Self {
+pub struct BlockHeaders<T>(pub Vec<T>);
+impl<T> From<Vec<T>> for BlockHeaders<T> {
```
mzabaluev: Same as above, add bound `T: property::Header`.

protocol-tokio/src/protocol/codec/message.rs
```diff
@@ -301,37 +314,42 @@ impl de::Deserialize for GetBlockHeaders {
...
-pub struct GetBlocks {
-    pub from: HeaderHash,
-    pub to: HeaderHash,
+pub struct GetBlocks<T> {
+    pub from: T,
+    pub to: T,
 }
-impl GetBlocks {
-    pub fn new(from: HeaderHash, to: HeaderHash) -> Self {
+impl<T> GetBlocks<T> {
```
mzabaluev: And here, `T: property::Block`.

protocol-tokio/src/protocol/codec/message.rs
```diff
@@ -351,63 +369,43 @@ impl de::Deserialize for GetBlocks {
...
-fn random_from<G: ::quickcheck::Gen>(g: &mut G) -> Vec<HeaderHash> {
-    let num: usize = ::quickcheck::Arbitrary::arbitrary(g);
-    ::std::iter::repeat_with(|| random_headerhash(g))
-        .take(num)
-        .collect()
-}
-
 #[cfg(test)]
-impl ::quickcheck::Arbitrary for GetBlockHeaders {
+impl<T: ::quickcheck::Arbitrary> ::quickcheck::Arbitrary for GetBlockHeaders<T> {
```
mzabaluev: Move these under `mod test`. It does not matter where in the crate an impl is defined and organizationally they belong there.

network-ntt/src/server.rs
```diff
@@ -0,0 +1,270 @@
...
+        })
+}
+
+/// Connect to another client.
+/// `run_connection` is spawned on the single heavyweight
+/// connection.
+pub fn connect<N: 'static, F>(
+    sockaddr: SocketAddr,
+    node: Server<N>,
+) -> impl future::Future<Item = impl futures::future::Future<Item=(),Error=()>, Error = ()>
```
mzabaluev: So, `accept` and `connect` return different anonymous `Future` types backed by the protocol connection. Does it mean that we have asymmetric protocol connections in NTT after all?

protocol-tokio/examples/client.rs
```diff
@@ -30,7 +30,10 @@ fn main() {
             Connection::connect(stream)
         })
         .and_then(|connection| {
-            let (sink, stream) = connection.split();
+            let (sink, stream): (
+                OutboundSink<_, u8, u8, u8, u8>,
```
mzabaluev: This is not very illustrative as an example. Maybe mockchain types should help here. It will also serve to verify soundness of the type work.

network-ntt/src/server.rs
```diff
@@ -0,0 +1,270 @@
...
+where
+    N: server::Node + Clone,
+    F: future::Future<Item = (), Error = ()>,
+    <<N as server::Node>::BlockService as server::block::BlockService>::Block : cbor_event::de::Deserialize,
```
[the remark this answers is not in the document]
qnikst: We need that cbor serialization to be backwards compatible with old protocol, as far as I understood chain-core does to guarantee, and I'm not sure that can implement exactly the same serialization. If I'm wrong I'll happily move to those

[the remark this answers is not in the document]
qnikst: We need that cbor serialization to be backwards compatible with old protocol, as far as I understood chain-core does to guarantee, and I'm not sure that can implement exactly the same serialization. If I'm wrong I'll happily move to those

network-core/src/server/block.rs
```diff
@@ -63,6 +72,18 @@ pub trait HeaderService {
     /// The future resolves to a stream that will be used by the protocol
     /// implementation to produce a server-streamed response.
     type GetHeadersFuture: Future<Item = Self::GetHeadersStream, Error = BlockError>;
+
+    // TODO: return a stream instead of the vector.
```
mzabaluev: Looks like it does return a stream already?

PR #404, opened 2019-01-22
Genesis generation

edolstra:
This is for https://github.com/input-output-hk/jormungandr/issues/35. It adds a program (currently in exe-common/examples, but should probably be moved somewhere else) that generates a genesis data file (in canonicalized JSON format), as well as private keys for the nodes and some initial non-AVVM balances. See 0c920f89b614dbaf33165343004f9324c9dc32e2 for usage info.

There are currently a bunch of genesis fields missing because they're not represented in `config::GenesisData`. I once added them to `config::GenesisData` as a `ChainParameters` struct in the unmerged PR #300, so if desired I could move that over to here (i.e. minus the update handling). However we're not currently using them anywhere.

There's also some miscellaneous cleanup to `Coin` and `Milli` (to catch some potential bugs) and to `BlockVersion` (cherry-picked from PR #300, but in retrospect not really needed here...).

### DIFF
--- a/cardano-c/src/transaction.rs
+++ b/cardano-c/src/transaction.rs
@@ -106,10 +106,10 @@ pub extern "C" fn cardano_transaction_builder_fee(tb: TransactionBuilderPtr) -> 
     let fee_algo = LinearFee::default();
 
     if let Ok(fee) = builder.calculate_fee(&fee_algo) {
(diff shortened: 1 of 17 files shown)


vincenthz:
one thing I've not followed properly is why most fields in the update are now commented ?

PR #413, opened 2019-01-25
network-core: Make client::Error more informative

mzabaluev:
The causes enumeration is now `ErrorKind` and the `Error` struct contains it
as a property. It also contains a chained error from the implementation.

### DIFF
--- a/network-core/src/client/error.rs
+++ b/network-core/src/client/error.rs
@@ -1,0 +1,51 @@
+use std::{error, fmt};
+
+/// Represents errors that can be returned by the node client implementation.
+#[derive(Debug)]
+pub struct Error {
+    kind: ErrorKind,
+    source: Box<dyn error::Error + Send + Sync>,
+}
+
+/// A list of general causes of client request errors.
+///
+/// This list is intended to grow over time and it is not recommended to
+/// exhaustively match against it.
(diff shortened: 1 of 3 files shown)


network-core/src/client/error.rs
```diff
@@ -0,0 +1,51 @@
...
+    fn source(&self) -> Option<&(dyn error::Error + 'static)> {
+        Some(self.source.as_ref())
+    }
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self.kind {
+            ErrorKind::Format => write!(f, "malformed payload received"),
+            ErrorKind::Rpc => write!(f, "protocol error"),
```
qnikst: can we ask for display for the source and write it as well? Otherwise information may be lost

mzabaluev: The source error is available through the `Error` trait, and in cardano-cli we do nice error printout that goes over the error chain. That printout would then redundantly print the underlying message.
That said, I've seen the other style and also used it in other error types, particularly when there is an enum with error value variants to match. No hard preference here.

qnikst: Thanks for the clarification, this approach works perfectly well for me then.

PR #425, opened 2019-01-31
add leadership algo for the mockchain

NicolasDP:


chain-impl-mockchain/src/leadership/bft.rs
```diff
@@ -0,0 +1,147 @@
...
+pub struct BftLeaderSelection<LeaderId> {
+    my: Option<BftRoundRobinIndex>,
+    leaders: Vec<LeaderId>,
+
+    current_leader: LeaderId,
+}
+
+#[derive(Debug)]
+pub enum Error {
+    BlockHasInvalidLeader(PublicKey, PublicKey),
```
mzabaluev: It would be more informative to have a structure with named fields, but this will do for the mockchain.

chain-impl-mockchain/src/leadership/bft.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// get the party leader id elected for a given slot
+    #[inline]
+    pub fn get_leader_at(&self, slotid: u64) -> &LeaderId {
+        let BftRoundRobinIndex(ofs) = self.offset(slotid);
+        &self.leaders[ofs]
+    }
+
+    /// check if this party is elected for a given slot
+    #[inline]
+    pub fn am_leader_at(&self, slotid: u64) -> IsLeading {
```
mzabaluev: Is this type supposed to represent both the set of leaders and the selected leader? The name also confusingly differs from `LeaderSelection::is_leader_at`.

chain-impl-mockchain/src/leadership/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+    type Error = Error;
+
+    fn diff(&self, input: &Self::Block) -> Result<Self::Update, Self::Error> {
+        let mut update = <Self::Update as property::Update>::empty();
+
+        match self {
+            LeaderSelection::BFT(ref bft) => {
+                update.bft = property::LeaderSelection::diff(bft, input).map_err(Error::Bft)?;
+            }
+            LeaderSelection::Genesis => {}
```
mzabaluev: Should there really be `unimplemented!()` so that it does not silently do nothing when we hit this branch in production?

chain-impl-mockchain/src/leadership/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        Ok(update)
+    }
+    fn apply(&mut self, update: Self::Update) -> Result<(), Self::Error> {
+        match self {
+            LeaderSelection::BFT(ref mut bft) => {
+                property::LeaderSelection::apply(bft, update.bft).map_err(Error::Bft)?;
+            }
+            LeaderSelection::Genesis => {}
```
mzabaluev: Same as above.

chain-impl-mockchain/src/leadership/bft.rs
NicolasDP: good point, and it overlaps with the method exposed by the `LeaderSelection` trait. I'll make it private and change the name too.

chain-impl-mockchain/src/leadership/mod.rs
NicolasDP: good point

chain-impl-mockchain/src/leadership/mod.rs
NicolasDP: good point

PR #430, opened 2019-02-01
Extend Header property with Id information.

qnikst:
Provide additional properties for the Header, so we
can retrieve block id from the header. This API may
be needed if blockchain implementation uses Headers
in the protocol instead of ID's.

### DIFF
--- a/chain-core/src/property.rs
+++ b/chain-core/src/property.rs
@@ -91,11 +91,17 @@ pub trait Block: Serialize + Deserialize {
 /// block's metadata via a network protocol or in other uses where the
 /// full content of the block is too bulky and not necessary.
 pub trait HasHeader {
+    /// The block header id.
+    type Id: BlockId;
+
     /// The block header type.
     type Header: Header;
 
     /// Retrieves the block's header.
     fn header(&self) -> Self::Header;
+
+    /// Retrieves the block's header id.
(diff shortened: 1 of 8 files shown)


chain-core/src/property.rs
```diff
@@ -91,11 +91,17 @@ pub trait Block: Serialize + Deserialize {
...
+    type Id: BlockId;
+
     /// The block header type.
     type Header: Header;
 
     /// Retrieves the block's header.
     fn header(&self) -> Self::Header;
+
+    /// Retrieves the block's header id.
+    fn id(&self) -> Self::Id;
```
mzabaluev: This seems redundant as the same method exists on trait `Block`. Instead, I think the `Header` trait should have this method.

qnikst: but the header is smaller than the block, doesn't it? The idea is that we want to get an `id` having only Header, but not the block itself.
At the same time `Block` trait doesn't imply `Header` trait so header trait method is not enough as well.

So as I see we need both, but maybe I should rename the method for header?

NicolasDP: why can't the trait `Header` provide the `id` method?

have it duplicated on both trait for now (Block and Header). This is easier.

qnikst:
@mzabaluev I've got your idea, fixed my solution in another PR

PR #433, opened 2019-02-04
network-ntt client implementation.

qnikst:


network-ntt/src/client.rs
```diff
@@ -0,0 +1,297 @@
...
+    channel: mpsc::UnboundedSender<Request<T>>,
+    phantom: PhantomData<Tx>,
+}
+
+/// Connect to the remote client. Returns future that can
+/// be run on any executor.
+pub fn connect<B, Tx>(
+    sockaddr: SocketAddr,
+) -> impl Future<
+    Item = (impl Future<Item = (), Error = ()>, ClientHandle<B, Tx>),
```
mzabaluev: This future result type looks a bit loaded. Could `ClientHandle` implement the lifecycle future? If not, make the item type a structure that implements the connection future and has a named accessor method for the handle.

network-ntt/src/client.rs
```diff
@@ -0,0 +1,297 @@
...
+
+pub struct StreamOnce<K>(oneshot::Receiver<K>);
+
+impl<T> Stream for StreamOnce<T> {
+    type Item = T;
+    type Error = core_client::Error;
+
+    fn poll(&mut self) -> Poll<Option<T>, Self::Error> {
+        match self.0.poll() {
+            _ => unimplemented!(),
```
mzabaluev: What should this do?

protocol-tokio/src/protocol/inbound_stream.rs
```diff
@@ -74,8 +80,8 @@ impl<T: AsyncRead, B: property::Block + property::HasHeader, Tx: property::Trans
 where
     B: cbor_event::Deserialize,
     B: cbor_event::Serialize,
-    B::Id: cbor_event::Deserialize,
-    B::Id: cbor_event::Serialize,
+    <B as property::Block>::Id: cbor_event::Deserialize,
+    <B as property::Block>::Id: cbor_event::Serialize,
```
mzabaluev: Might as well put both bounds on one line joined with a `+`, here and in other places below.

mzabaluev: Are these changes a remnant of having the `Id` associated type on `HasHeader`? This disambiguation should not be needed.

network-ntt/src/client.rs
qnikst: I'm not sure, the problem here is the structure of the program that we want to support.
My idea was that we can spawn a future, after that we can keep a handle in order to run communication. The problem with communication is that we should be listening and replying to the messages even we don't send any request. So I decided to separate concerns and return future that can be spawned in any meaningful way and a communication handle.

network-ntt/src/client.rs
qnikst: Oh, this guy was for the streams, but we don't have API that maps on the n-t-t API nicely, I'll see if I can implement `pull_blocks_to_tip`, then I'll fill this one.

network-ntt/src/client.rs
mzabaluev: My main concern here is stylistic: the anonymous tuple that does not provide clues how to use it or what its members are. I think it should be a named struct and encapsulate both the future and the handle. If the future type is too cumbersome to spell out explicitly, it's OK to leave it as is.

network-ntt/src/client.rs
```diff
@@ -247,6 +247,42 @@ enum Command<B: Block + HasHeader, Tx: TransactionId> {
...
+    A1(A1),
+    A2(A2),
+    A3(A3),
+    A4(A4),
+    A5(A5),
+    A6(A6),
+    A7(A7),
+}
+
+impl<A1,A2,A3,A4,A5,A6,A7> Future for V<A1, A2, A3, A4,A5,A6,A7>
```
mzabaluev: I'd be surprised if rustfmt is silent on uneven spacing here :)

mzabaluev:
Accepting "with reservations" that I'm trying to resolve in another branch. We need changes to chain-core and other things done here in order to continue with jormungandr update.

PR #434, opened 2019-02-05
remove the type parameter of the trait Ledger and of HasTransaction

### LINKED ISSUE #417: change the HasTransaction trait
It is a bit of a confusing trait right now, we might want to look at simplifying it a little here. Maybe even create additional traits to cover the different accessors we may need.

NicolasDP:
This will simplify the code a bit in the sense it will make it simpler to manipulate trait only object.

close #417

### DIFF
--- a/chain-core/src/property.rs
+++ b/chain-core/src/property.rs
@@ -128,9 +128,11 @@ pub trait Transaction: Serialize + Deserialize {
 /// * UTxO
 /// * certificate registrations
 /// * ...
-pub trait HasTransaction<T: Transaction> {
+pub trait HasTransaction {
+    type Transaction: Transaction;
+
     /// returns an iterator over the Transactions
-    fn transactions<'a>(&'a self) -> std::slice::Iter<'a, T>;
+    fn transactions<'a>(&'a self) -> std::slice::Iter<'a, Self::Transaction>;
 }
 
 /// Updates type needs to implement this feature so we can easily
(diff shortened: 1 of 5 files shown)


chain-core/src/property.rs
```diff
@@ -156,7 +158,7 @@ pub trait Update {
 /// Define the Ledger side of the blockchain. This is not really on the blockchain
 /// but should be able to maintain a valid state of the overall blockchain at a given
 /// `Block`.
-pub trait Ledger<T: Transaction> {
+pub trait Ledger: Sized {
```
mzabaluev: It's odd that the `Sized` bound has to be included. I can't see why, but it likely does not hurt.

chain-core/src/property.rs
```diff
@@ -370,11 +380,10 @@ pub mod testing {
 
     /// Generate a number of transactions and run them, it's not
     /// expected to have any errors during the run.
-    pub fn run_valid_transactions<'a, G, L, T>(g: &'a mut G, ledger: &'a mut L, n: usize) -> ()
+    pub fn run_valid_transactions<'a, G, L>(g: &'a mut G, ledger: &'a mut L, n: usize) -> ()
```
mzabaluev: The lifetime bound is probably redundant here.

chain-core/src/property.rs
NicolasDP: It is for a following PR I will make 🤦‍♂️ sorry for including this too early.

PR #437, opened 2019-02-12
`sync.rs` - tiny refactoring and simplification

vantuz-subhuman:
Just simplified some code a very little

### DIFF
--- a/exe-common/src/sync.rs
+++ b/exe-common/src/sync.rs
@@ -42,22 +42,19 @@ fn net_sync_to<A: Api>(
         "Network TIP is       : {} ({}) <- {}",
         tip.hash,
         tip.date,
-        tip_header.get_previous_header()
+        tip.parent
     );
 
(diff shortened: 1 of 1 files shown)


exe-common/src/sync.rs
```diff
@@ -167,16 +164,16 @@ fn net_sync_to<A: Api>(
...
         genesis_data,
-        if our_tip.1 {
-            &our_tip.0.parent
+        if our_tip_is_genesis {
+            &our_tip.parent
         } else {
-            &our_tip.0.hash
+            &our_tip.hash
         },
     )?;
```
vantuz-subhuman: Just to be extra sure - can someone please clarify why the **parent** is used here if the tip is a genesis? 🤔Is this "genesis" refers to EBB?

vantuz-subhuman: Ah yes, seems obvious to be EBB, since `BlockDate::Boundary` is used. My bad.

PR #438, opened 2019-02-12
add the SignedBlockSummary

NicolasDP:
this will act as a Block Header like in Cardano except that it is only
extracted from the data of the SignedBlock. This is because there is
no such thing as block header in the mock chain (as this is completely
not necessary).

### DIFF
--- a/chain-impl-mockchain/src/block.rs
+++ b/chain-impl-mockchain/src/block.rs
@@ -36,6 +36,26 @@ pub struct SignedBlock {
     pub block: Block,
 }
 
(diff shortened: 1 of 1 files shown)


chain-impl-mockchain/src/block.rs
```diff
@@ -32,10 +32,31 @@ pub struct SignedBlock {
     pub public_key: PublicKey,
     /// List of cryptographic signatures that verifies the block.
     pub signature: Signature,
+
```
NicolasDP: arf, going to remove this line from the commig

PR #440, opened 2019-02-12
mockchain: Add Display impl for BlockDate

mzabaluev:


chain-impl-mockchain/src/key.rs
```diff
@@ -116,6 +118,12 @@ impl property::Deserialize for Hash {
 impl property::BlockId for Hash {}
 impl property::TransactionId for Hash {}
 
+impl fmt::Display for Hash {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
```
NicolasDP: oops: #441

mzabaluev: Rebased to only add the `BlockDate` impl.

PR #443, opened 2019-02-13
Improve trait bounds on blockchain entities

mzabaluev:
Add `chain-core` property trait `FromStr` to impose extra requirements on parsing values such as block dates from strings.

Require the error types for deserialization and string parsing to be safe for passing between threads, so that they can be boxed and contained in e.g. `client::Error` of `network-core`.

Trait work in gRPC client code to reduce repetitive bound clauses.

### DIFF
--- a/chain-core/src/property.rs
+++ b/chain-core/src/property.rs
@@ -311,6 +311,11 @@ pub trait Deserialize: Sized {
     fn deserialize<R: std::io::BufRead>(reader: R) -> Result<Self, Self::Error>;
 }
 
+/// Defines the way to parse the object from a UTF-8 string.
+///
+/// This is like the standard `FromStr` trait, except that it imposes
+/// additional bounds on the error type to make it more usable for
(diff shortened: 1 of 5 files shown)


chain-core/src/property.rs
```diff
@@ -307,11 +306,29 @@ pub trait Serialize {
...
+where
+    T: std::str::FromStr,
+    <T as std::str::FromStr>::Err: std::error::Error + Send + Sync + 'static,
+{
+    type Error = <T as std::str::FromStr>::Err;
+
+    fn from_str(s: &str) -> Result<Self, Self::Error> {
+        std::str::FromStr::from_str(s)
+    }
+}
```
NicolasDP: argh... this is very hacky... :/

mzabaluev: https://github.com/rust-lang/rust/issues/32722 does not allow just adding a subtrait with extra bounds on `FromStr::Err`.

PR #445, opened 2019-02-15
Fixed input/output comparison

vantuz-subhuman:


NicolasDP:
@vantuz-subhuman was it correct at the end?

vantuz-subhuman:
I mixed up, the check is correct, it adds an error if output+fee is LARGER, which is correct. I looked at the comment just above it and confused the comparison =) Apologies

vantuz-subhuman:
Here's the actual issue found: https://github.com/input-output-hk/rust-cardano/pull/446

PR #446, opened 2019-02-15
Fixed genesis block parsing to divide fee policy correctly

vantuz-subhuman:


vantuz-subhuman:
This is how fee values are specified in the mainnet genesis:

image

They are way too big, to be just divided by 1000. The actual size is `lovelaces * 1'000'000'000`, and for multiplier we calculate:
```
1) 43'946'000'000 / 1'000'000'000
2) (43'946'000'000 % 1'000'000'000) / 1'000'000
```

Which gives us tuple of: `(43, 946)` - exactly the correct values for `Milli::new`

vantuz-subhuman:
If some genesis will ever have meaningful last 6 digits (e.g. `43'946'123'000`) - this information will be lost, but then it means that type `Milli` is not suitable to process this larger information, and will have to be replaces with something like `Nano` )

edolstra:
Thanks @vantuz-subhuman! I have gone for a slightly simpler fix in #449.

PR #447, opened 2019-02-15
Introduce tip_header method in the HeaderService.

qnikst:


network-grpc/src/client.rs
```diff
@@ -385,13 +385,19 @@ where
 
 impl<T, S, E> HeaderService<T> for Client<S, E>
 where
-    T: HasHeader,
```
mzabaluev: Is this extra bound necessary?

network-grpc/src/client.rs
```diff
@@ -385,13 +385,19 @@ where
...
+    T: Block + HasHeader,
```
NicolasDP: this is not necessarily true (that a block has a header).

Why not parameterising the trait with a `H: Header` parameter ?

qnikst: Fixed, it appeared to be a redundant change, left after previous work.

network-grpc/src/client.rs
qnikst: actually not, fixed

qnikst:
Please take another look

PR #448, opened 2019-02-15
Further refine abstract iteration API in chain-core

mzabaluev:
The `Option` return of `HasTransaction::transactions()` is unwieldy.
It's better to have an empty array in the implementation type that does
not have transactions, and iterate over the slice reference.

The iterable accessor types do not have to be sized, so lift the sized
bound from these associated types.

### DIFF
--- a/cardano/src/block/block.rs
+++ b/cardano/src/block/block.rs
@@ -364,7 +364,7 @@ impl chain_core::property::HasTransaction for Block {
     type Transactions = [TxAux];
     fn transactions(&self) -> &Self::Transactions {
         match self {
-            Block::BoundaryBlock(blk) => &blk.body.no_transactions,
+            Block::BoundaryBlock(_) => &[],
             Block::MainBlock(blk) => &blk.body.tx,
         }
     }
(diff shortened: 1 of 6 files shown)


cardano/src/block/block.rs
```diff
@@ -361,11 +361,11 @@ impl chain_core::property::Deserialize for Block {
...
 
 impl chain_core::property::HasTransaction for Block {
-    type Transactions = Vec<TxAux>;
-    fn transactions(&self) -> Option<&Self::Transactions> {
+    type Transactions = [TxAux];
+    fn transactions(&self) -> &Self::Transactions {
         match self {
-            Block::BoundaryBlock(_) => None,
-            Block::MainBlock(ref blk) => Some(&blk.body.tx),
+            Block::BoundaryBlock(blk) => &blk.body.no_transactions,
```
NicolasDP: can't we return a static ref here `& []` instead of adding a un-necessary and confusing field to the boundary block ?

mzabaluev: Ah, of course!

PR #450, opened 2019-02-15
network-transport fixes

qnikst:


network-grpc/src/client.rs
```diff
@@ -385,13 +385,19 @@ where
 
 impl<T, S, E> HeaderService<T> for Client<S, E>
 where
-    T: HasHeader,
+    T: Block + HasHeader,
```
mzabaluev: Please rebase this after #447 is merged, it would be easier to review just the relevant changes here.

network-ntt/src/client.rs
```diff
@@ -171,7 +210,7 @@ pub trait NttBlock<D, I, H>:
 where
     D: NttDate,
     I: NttId,
-    H: NttHeader<D, I>,
+    H: NttHeader<D, I> + Clone + core::fmt::Debug,
```
mzabaluev: These bounds are already implied by `NttHeader`?

network-ntt/src/client.rs
```diff
@@ -241,35 +288,26 @@ impl<B: Block + HasHeader> ConnectionState<B> {
...
     A2(A2),
     A3(A3),
     A4(A4),
     A5(A5),
-    A6(A6),
-    A7(A7),
 }
 
-impl<A1, A2, A3, A4, A5, A6, A7> Future for V<A1, A2, A3, A4, A5, A6, A7>
+impl<A1, A2, A3, A4, A5> Future for V<A1, A2, A3, A4, A5>
```
mzabaluev: This horror should be removed by my state machine rewrite, but let it linger for a the time being.

protocol-tokio/src/protocol/codec/handshake.rs
```diff
@@ -17,7 +17,7 @@ use super::{MessageCode, MessageType};
 pub struct ProtocolMagic(u32);
 impl Default for ProtocolMagic {
     fn default() -> Self {
-        ProtocolMagic(764824073)
+        ProtocolMagic(1097911063)
```
mzabaluev: I wonder if we could have other `ProtocolMagic` values, or is the whole thing just an unnecessarily glorified const?

protocol-tokio/src/protocol/codec/message.rs
```diff
@@ -176,34 +173,12 @@ where
     }
 
     pub fn expect_bytes(event: nt::Event) -> Result<Self, nt::Event> {
+        println!("[{}:{}] expect bytes: data {:?}", file!(), line!(), event);
```
mzabaluev: Debug `println`s detected! Please remove before merging.

protocol-tokio/src/protocol/codec/message.rs
```diff
@@ -176,34 +173,12 @@ where
...
-            MessageType::MsgBlock => {
-                Ok(Message::Block(lwcid, cbor.deserialize_complete().unwrap()))
-            }
-            MessageType::MsgSubscribe1 => {
-                let v: u64 = cbor.deserialize_complete().unwrap();
-                let keep_alive = v == 43;
-                Ok(Message::Subscribe(lwcid, keep_alive))
-            }
-            _ => unimplemented!(),
+        println!("[{}:{}] expect bytes: {:?}", file!(), line!(), lwcid);
```
mzabaluev: Same as above.

protocol-tokio/src/protocol/codec/handshake.rs
NicolasDP: we should have only one (maybe to) const protocol magic, mainnet and staging
the other ones are testing networks.

NicolasDP: add a comment that this should be replaced by a configuration/setting mechanism...


```rust
// TODO: get this from user input
```

and for now I prefer the default to be mainnet's protocol magic.

mzabaluev: I think it would be less error-prone to define the magic in the code that implements the protocol, rather than allow it to be redefined with an arbitrary value.

mzabaluev: So this is not just the protocol magic, but also determines the chain.

mzabaluev: Upon reading the discussion above, I agree that it should be obtained from the same source that defines the genesis block hash.

qnikst: We should take that value as a parameter, I think I'll need to extend network configuration, so this value will be passed by the user. I'd like to do that in the separate commit unless it's a blocker.

network-ntt/src/client.rs
qnikst: Afaik the commit that introduced the change didn't compile yet? Otherwise I will happily rebase on that.

PR #451, opened 2019-02-15
Allow network-ntt implementation backend in exe-commons

qnikst:


exe-common/src/config.rs
```diff
@@ -49,13 +49,16 @@ pub mod net {
...
         Http(String),
+        Ntt(String),
     }
     impl Peer {
         /// analyse the content of the given `addr` and construct the correct kind
         /// of `Peer` accordingly.
         pub fn new(addr: String) -> Self {
             if addr.starts_with(r"http://") || addr.starts_with(r"https://") {
                 Peer::http(addr)
+            } else if addr.starts_with(r"ntt://") {
```
mzabaluev: Eek, vendors should not invent URI schemes just like that. Use `x-cardano-ntt` perhaps.

exe-common/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![type_length_limit = "2097152"]
```
mzabaluev: I wonder what monster type causes this.
What's the error if this directive is removed?

exe-common/src/network/ntt.rs
```diff
@@ -0,0 +1,73 @@
...
+use network_core::client::block::HeaderService;
+use network_ntt::client as ntt;
+use std::net::SocketAddr;
+
+use cardano::{
+    block::{Block, BlockHeader, HeaderHash, RawBlock},
+    tx::{TxAux, TxId},
+};
+
+pub struct NetworkCore {
```
mzabaluev: Why is this named `NetworkCore`?

exe-common/src/network/ntt.rs
```diff
@@ -0,0 +1,73 @@
...
+                    println!("Exited");
+                }));
+                Ok(NetworkCore { handle, rt })
+            }
+            Err(_err) => unimplemented!(),
+        }
+    }
+}
+
+impl Api for NetworkCore {
```
mzabaluev: Ugh, why do we have this trait named `Api` to begin with. Not a problem of this PR, of course.

exe-common/src/network/ntt.rs
```diff
@@ -0,0 +1,73 @@
...
+    fn get_tip(&mut self) -> Result<BlockHeader> {
+        self.handle
+            .tip_header()
+            .map_err(|e| Error::from(std::io::Error::new(std::io::ErrorKind::Other, e)))
+            .wait()
+    }
+
+    fn wait_for_new_tip(&mut self, _prev_tip: &HeaderHash) -> Result<BlockHeader> {
+        panic!("not yet ready")
```
mzabaluev: Could be just `unimplemented!()`

exe-common/src/network/peer.rs
```diff
@@ -29,6 +31,16 @@ impl Peer {
             config::net::Peer::Http(addr) => {
                 Ok(Peer::Http(hermes::HermesEndPoint::new(addr, network)))
             }
+            config::net::Peer::Ntt(addr) => {
+                let mut addrs_iter = addr.to_socket_addrs().unwrap();
+                match addrs_iter.next() {
+                    Some(addr) => ntt::NetworkCore::new(addr).map(Peer::Ntt),
+                    None => Err(Error::from(std::io::Error::new(
+                        std::io::ErrorKind::Other,
```
mzabaluev: I would say `ErrorKind::InvalidInput` fits better here. But even better is to have a case for this in `network::Error` rather than go through `std::io::Error`.

network-grpc/src/client.rs
```diff
@@ -1,6 +1,6 @@
 use crate::gen::{self, node::client as gen_client};
 
-use chain_core::property::{self, Deserialize, FromStr, HasHeader, Serialize};
+use chain_core::property::{self, Block, Deserialize, FromStr, HasHeader, Serialize};
```
mzabaluev: Rebase this as well to eliminate cross-pollination from other branches. I'll continue reviewing from the clean branch.

exe-common/src/config.rs
NicolasDP: that's a good point. Actually what about using multiaddr?

exe-common/src/lib.rs
NicolasDP: holy 🐮 ! we have a type that needs 2 MB ?

exe-common/src/network/ntt.rs
NicolasDP: old exe-common things...

exe-common/src/network/peer.rs
qnikst: I'm not sure I want to propagate all the errors in the rust-cardano, I'm not sure I'm ready for that endeavour.

exe-common/src/network/ntt.rs
qnikst: I was planning to have a single structure that will keep runtime for all the threads, but that didn't work out. Any better name suggestion will be appreciated.

exe-common/src/lib.rs
qnikst: Hurray after a number of rewrites of the ntt-client, this requirement has gone.

qnikst:
Please take another look, I've addressed all the comments except for naming and possibly multiaddr.
Currently the only problem I see is that I'm creating a runtime for each connection, ideally, I should create only one runtime for all the program, but I'm not sure how to fit it here.

exe-common/src/network/peer.rs
mzabaluev: In fact, what's needed here is handling the error return of `to_socket_addrs()` above instead of dying on an invalid address.

Consider adding a variant `InvalidPeerAddress(String)` to `network::Error`; the address string that failed parsing goes inside the value.

protocol-tokio/src/protocol/codec/handshake.rs
```diff
@@ -43,6 +43,12 @@ impl de::Deserialize for ProtocolMagic {
     }
 }
 
+impl From<&u32> for ProtocolMagic {
```
mzabaluev: Why does it need to be a reference? `u32` is copyable.

qnikst: Fixed

exe-common/src/network/peer.rs
qnikst: Fixed

exe-common/src/config.rs
```diff
@@ -49,13 +49,16 @@ pub mod net {
...
+        Ntt(String),
     }
     impl Peer {
         /// analyse the content of the given `addr` and construct the correct kind
         /// of `Peer` accordingly.
         pub fn new(addr: String) -> Self {
             if addr.starts_with(r"http://") || addr.starts_with(r"https://") {
                 Peer::http(addr)
+            } else if addr.starts_with(r"x-cardano-ntt://") {
+                Peer::ntt(addr[17..].to_string())
```
mzabaluev: Could use .len() instead of counting bytes.

exe-common/src/network/peer.rs
```diff
@@ -29,6 +31,13 @@ impl Peer {
             config::net::Peer::Http(addr) => {
                 Ok(Peer::Http(hermes::HermesEndPoint::new(addr, network)))
             }
+            config::net::Peer::Ntt(addr) => {
+                let mut addrs_iter = addr.to_socket_addrs().unwrap();
```
mzabaluev: The `unwrap` is still there, meaning that jormungandr will panic on any invalid address?

qnikst: I hope this code is never used in jormundandr.. but you are right that cardano-cli will fail

mzabaluev: Sorry, I meant cardano-cli. It should be easy to check for the error here and convert it to the same `InvalidPeerAddress` error variant.

exe-common/src/network/error.rs
```diff
@@ -66,6 +67,7 @@ impl fmt::Display for Error {
             Error::NoSuchBlock(hash) => write!(f, "Requested block {} does not exist", hash),
             Error::StorageError(_) => write!(f, "Storage error"),
             Error::BlockError(_) => write!(f, "Block error"),
+            Error::InvalidPeerAddress(addr) => write!(f, "Block error {}", addr),
```
mzabaluev: Something like `"Invalid peer address '{}'", addr`, perhaps?

exe-common/src/config.rs
qnikst: fixed

exe-common/src/network/peer.rs
qnikst: Fixed

exe-common/src/network/error.rs
qnikst: Oh.. my mistake, thanks, fixed.

exe-common/src/network/peer.rs
mzabaluev: Hmm, the `unwrap` is still there?

exe-common/src/config.rs
```diff
@@ -49,13 +49,16 @@ pub mod net {
...
+        Ntt(String),
     }
     impl Peer {
         /// analyse the content of the given `addr` and construct the correct kind
         /// of `Peer` accordingly.
         pub fn new(addr: String) -> Self {
             if addr.starts_with(r"http://") || addr.starts_with(r"https://") {
                 Peer::http(addr)
+            } else if addr.starts_with(r"x-cardano-ntt://") {
+                Peer::ntt(addr["x-cardano-ntt://".len()..].to_string())
```
mzabaluev: OK, though using a string constant would be better.

mzabaluev: It's also not an `r`-literal, though I'm not sure there is a practical difference. Small things...

mzabaluev:
This is good to go after a touch of rustfmt.

PR #452, opened 2019-02-15
Sync / OBFT-ready

vantuz-subhuman:
Tried to change all the places in the sync and related code where it were strictly dependent on checking boundary blocks. All changes made in such a way that now it supports processing epochs with or without EBB.

⚠️**Missing:**
1. OBFT-specific block validation (slot-leader)
2. Validation of OBFT itself. E.g. checking if epochs before OBFT **DO** have EBBs, and checking that epochs after OBFT - **DON'T**


NicolasDP:
our first github draft \o/

exe-common/src/sync.rs
```diff
@@ -143,13 +143,17 @@ fn net_sync_to<A: Api>(
         // Iterate to the last block in the previous epoch.
         let mut cur_hash = our_tip.hash.clone();
         loop {
-            let block_raw = storage.read_block(&cur_hash.into()).unwrap();
+            let block_raw = storage.read_block(&cur_hash.clone().into()).unwrap();
```
vantuz-subhuman: I had to add `.clone()` here, because `cur_hash` reassignment became optional, and compiler were complaining that it might be same already moved value. I figured clone is ok, since hash is a very simple value, but I might be doing something stupid from Rust POV here.

mzabaluev:
Please rebase the branch to remove commits from elsewhere that have already been merged, so that it's easier to review.

PR #453, opened 2019-02-17
feature / Block-by-height

vantuz-subhuman:
Added some new functionality to support querying blocks by chain-height

### DIFF
--- a/cardano/src/block/date.rs
+++ b/cardano/src/block/date.rs
@@ -65,14 +65,14 @@ impl BlockDate {
             &BlockDate::Normal(ref s) => s.epoch,
         }
     }
-    pub fn get_slotid(&self) -> Option<SlotId> {
+    pub fn slotid(&self) -> Option<SlotId> {
         match self {
             &BlockDate::Boundary(_) => None,
             &BlockDate::Normal(ref s) => Some(s.slotid),
         }
     }
-    pub fn get_epoch_and_slot(&self) -> (EpochId, Option<SlotId>) {
-        (self.get_epochid(), self.get_slotid())
+    pub fn epoch_and_slot(&self) -> (EpochId, Option<SlotId>) {
+        (self.get_epochid(), self.slotid())
     }
     pub fn next(&self) -> Self {
         match self {
--- a/cardano/src/block/date.rs
+++ b/cardano/src/block/date.rs
@@ -65,6 +65,15 @@ impl BlockDate {
             &BlockDate::Normal(ref s) => s.epoch,
         }
     }
+    pub fn get_slotid(&self) -> Option<SlotId> {
+        match self {
+            &BlockDate::Boundary(_) => None,
+            &BlockDate::Normal(ref s) => Some(s.slotid),
+        }
+    }
+    pub fn get_epoch_and_slot(&self) -> (EpochId, Option<SlotId>) {
+        (self.get_epochid(), self.get_slotid())
+    }
     pub fn next(&self) -> Self {
         match self {
             &BlockDate::Boundary(e) => BlockDate::Normal(EpochSlotId {
--- a/cardano/src/block/types.rs
+++ b/cardano/src/block/types.rs
@@ -223,7 +223,7 @@ impl SscProof {
     }
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub struct ChainDifficulty(u64);
 
 impl fmt::Display for ChainDifficulty {
--- a/cardano/src/block/types.rs
+++ b/cardano/src/block/types.rs
@@ -223,7 +223,7 @@ impl SscProof {
     }
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Clone, Copy)]
 pub struct ChainDifficulty(u64);
 
 impl fmt::Display for ChainDifficulty {
--- a/cardano/src/block/types.rs
+++ b/cardano/src/block/types.rs
@@ -224,7 +224,7 @@ impl SscProof {
 }
 
 #[derive(Debug, Clone, Copy)]
-pub struct ChainDifficulty(pub u64);
+pub struct ChainDifficulty(u64);
 
 impl fmt::Display for ChainDifficulty {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
@@ -234,6 +234,11 @@ impl fmt::Display for ChainDifficulty {
 impl From<u64> for ChainDifficulty {
     fn from(f: u64) -> Self {
         ChainDifficulty(f)
+    }
+}
+impl From<ChainDifficulty> for u64 {
+    fn from(cd: ChainDifficulty) -> Self {
+        cd.0
     }
 }
 
--- a/cardano/src/block/types.rs
+++ b/cardano/src/block/types.rs
@@ -224,7 +224,7 @@ impl SscProof {
 }
 
 #[derive(Debug, Clone, Copy)]
-pub struct ChainDifficulty(u64);
+pub struct ChainDifficulty(pub u64);
 
 impl fmt::Display for ChainDifficulty {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
--- a/exe-common/src/sync.rs
+++ b/exe-common/src/sync.rs
@@ -374,6 +374,7 @@ fn finish_epoch(
         &storage,
         &packhash,
         epoch_id,
+        index,
         Some((chain_state, genesis_data)),
     );
 
--- a/exe-common/src/sync.rs
+++ b/exe-common/src/sync.rs
@@ -356,7 +356,7 @@ fn finish_epoch(
 ) -> Result<()> {
     let epoch_id = epoch_writer_state.epoch_id;
     let (packhash, index) = pack::packwriter_finalize(&storage.config, epoch_writer_state.writer);
-    let (lookup, tmpfile) = pack::create_index(&storage, &index, epoch_id as u32);
+    let (lookup, tmpfile) = pack::create_index(&storage, &index);
     tmpfile.render_permanent(&storage.config.get_index_filepath(&packhash))?;
     storage.add_lookup(packhash, lookup);
     let epoch_time_elapsed = epoch_writer_state.write_start_time.elapsed().unwrap();
@@ -373,7 +373,7 @@ fn finish_epoch(
     assert_eq!(chain_state.last_date.unwrap().get_epochid(), epoch_id);
 
     epoch::epoch_create(
-        &storage,
+        storage,
         &packhash,
         epoch_id,
         index,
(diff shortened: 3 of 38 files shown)


storage-units/src/indexfile.rs
```diff
@@ -50,16 +50,28 @@ fn offset_hashes(bloom_size: u32) -> u64 {
...
+    }
 }
 
 pub type IndexOffset = u32;
 
+#[derive(Clone, Debug)]
+pub enum IndexOffsetType {
+    Standard,
+    Unsorted,
+}
```
vantuz-subhuman: Discussed with @vincenthz - adding stuff to index files is undesirable at this moment, cuz of expected changes. Will change this whole logic a bit to store the unsorted offsets in this file: https://github.com/input-output-hk/rust-cardano/blob/master/storage/src/config.rs#L72 

This means that reading logic for unsorted offsets (un-indexed) will be very different from sorted offsets (indexed), and there will be no need for logic-switches like this one

storage-units/src/indexfile.rs
```diff
@@ -50,16 +50,28 @@ fn offset_hashes(bloom_size: u32) -> u64 {
...
+
 // The parameters associated with the index file.
 // * the bloom filter size in bytes
+// * the ordinal index for the lookup (e.g. epoch)
 pub struct Params {
     pub bloom_size: u32,
+    pub ordinal: u32,
```
vantuz-subhuman: 1. Is it ok to name this parameter like this?
2. Is it ok to put it in lookup params?

storage-units/src/indexfile.rs
```diff
@@ -149,10 +161,11 @@ impl Index {
         let bloom_size = default_bloom_size(entries);
         let params = Params {
             bloom_size: bloom_size,
+            ordinal: ordinal,
         };
 
         write_size(&mut hdr_buf[0..4], bloom_size);
-        write_size(&mut hdr_buf[4..8], 0);
+        write_size(&mut hdr_buf[4..8], ordinal);
```
vantuz-subhuman: It is ok to steal this unused but reserved 4-byte "field" in lookup header, or should I better add a new one. Note that extending the header size will require a change in reading logic in many places, if this field were prepared for something - maybe it might be better for me to take it now, cuz for you it should be way easier to add a new one and not mess something up later, when needed. I can add a new one myself too, but just thinking of risk/time handling )

storage-units/src/indexfile.rs
```diff
@@ -198,6 +211,14 @@ impl Index {
             write_offset(&mut buf, ofs);
             tmpfile.write_all(&buf[..])?;
         }
+
+        // Write ordered offsets to seek blocks by height
+        for ofs in self.offsets.iter() {
+            let mut buf = [0u8; OFF_SIZE];
+            write_offset(&mut buf, *ofs);
+            tmpfile.write_all(&buf[..])?;
+        }
```
vantuz-subhuman: This logic will be moved to another file (https://github.com/input-output-hk/rust-cardano/blob/master/storage/src/config.rs#L72), we will not extend the index field for now

storage/src/lib.rs
```diff
@@ -104,7 +110,7 @@ pub type Result<T> = result::Result<T, Error>;
 
 pub struct Storage {
     pub config: StorageConfig,
-    lookups: BTreeMap<PackHash, indexfile::Lookup>,
+    pub lookups: LinkedHashMap<PackHash, indexfile::Lookup>,
 }
```
vantuz-subhuman: 1. `BTreeMap` is replaced with `LinkedHashMap` in order to preserve the insertion order, cuz we need epoch indexes to be in the "chain order" for block-by-height query to work.
2. `pub` were added only for experiments/debug, I will remove it

storage/src/lib.rs
```diff
@@ -127,7 +133,7 @@ macro_rules! try_open {
 
 impl Storage {
     pub fn init(cfg: &StorageConfig) -> Result<Self> {
-        let mut lookups = BTreeMap::new();
+        let mut lookups = LinkedHashMap::with_capacity(1000);
```
vantuz-subhuman: 1000 is completely arbitrary here. `LinkedHashMap` implements auto-re-alloc on insert if out of capacity, so this one only ensures there will not be many re-allocations on the initial read. We have 73 epochs per year atm, so 1000 is way ahead of limitations. Maybe we can read the current number of existing indexes here first and init the map with this capacity plus some buffer.

storage/src/lib.rs
```diff
@@ -191,6 +200,19 @@ impl Storage {
...
+        let mut h = height;
+        for (packref, lookup) in self.lookups.iter() {
+            let indexfile::FanoutTotal(total) = lookup.fanout.get_total();
+            if h < (total as u64) {
+                return Ok(BlockLocation::Packed(packref.clone(), h as u32, indexfile::IndexOffsetType::Unsorted));
+            }
+            h -= total as u64;
+        }
+        Err(Error::BlockHeightNotFound(height))
+    }
```
vantuz-subhuman: This is the core logic for the block-by-height query search. Super simple, but for this to work - we need to:
1. Have lookups to be ordered in the chain order
2. Have block offsets available in the chain order for each epoch

exe-common/src/sync.rs
```diff
@@ -354,7 +354,7 @@ fn finish_epoch(
 ) -> Result<()> {
     let epoch_id = epoch_writer_state.epoch_id;
     let (packhash, index) = pack::packwriter_finalize(&storage.config, epoch_writer_state.writer);
-    let (lookup, tmpfile) = pack::create_index(&storage, &index);
+    let (lookup, tmpfile) = pack::create_index(&storage, &index, epoch_id as u32);
```
vantuz-subhuman: Gotta pass `epoch_id` to index creation here, to store it as `ordinal` parameter, to be able to sort lookups later.

cardano/src/block/types.rs
```diff
@@ -224,7 +224,7 @@ impl SscProof {
 }
 
 #[derive(Debug, Clone, Copy)]
-pub struct ChainDifficulty(u64);
```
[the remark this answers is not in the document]
vantuz-subhuman: Oh nice! Will do for sure!

storage-units/src/indexfile.rs
```diff
@@ -71,7 +73,7 @@ pub struct Lookup {
 pub struct Fanout([u32; FANOUT_ELEMENTS]);
 pub struct FanoutStart(u32);
 pub struct FanoutNb(pub u32);
-pub struct FanoutTotal(u32);
+pub struct FanoutTotal(pub u32);
```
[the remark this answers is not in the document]
vantuz-subhuman: Seems like something to be discussed then when we have opportunity 👍 I like some extractors more, like the `impl From`

exe-common/src/sync.rs
```diff
@@ -200,6 +200,8 @@ fn net_sync_to<A: Api>(
                 // yet. Instead we write this block to disk separately.
                 let block_hash = types::header_to_blockhash(&block_hash);
                 blob::write(storage, &block_hash, block_raw.as_ref()).unwrap();
+                // Add lose block to index
```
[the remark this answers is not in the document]
vantuz-subhuman: Fixed! 🙌

cardano/src/block/date.rs
```diff
@@ -65,6 +65,15 @@ impl BlockDate {
             &BlockDate::Normal(ref s) => s.epoch,
         }
     }
+    pub fn get_slotid(&self) -> Option<SlotId> {
```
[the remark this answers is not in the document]
vantuz-subhuman: Fixed! 👍 I like these names better too, but went for compatibility with `get_epochid` =\

storage-units/src/indexfile.rs
```diff
@@ -226,19 +232,34 @@ impl Lookup {
...
 }
 
+pub fn write_offsets_to_file<'a, I: Iterator<Item = &'a Offset>>(tmpfile: &mut TmpFile, offsets: I) -> Result<()> {
+    for ofs in offsets {
+        let mut buf = [0u8; OFF_SIZE];
+        write_offset(&mut buf, *ofs);
+        tmpfile.write_all(&buf[..])?;
+    }
+    Ok(())
+}
```
vantuz-subhuman: Maybe this function makes more sense in `serialize.rs`?

cardano/src/block/date.rs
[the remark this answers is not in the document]
mzabaluev: We should fix those names as well before we stabilize the API.

storage/src/lib.rs
```diff
@@ -320,19 +339,26 @@ impl Storage {
...
     }
 
+    /// Returns:
+    /// - Ok of Some entry - if found in loose index
+    /// - Ok of None - if requested height is low enough to not be in the index already
+    /// - Err of `BlockHeightNotFound` - if requested height is too large and out of index bounds
     pub fn get_from_loose_index(
         &self,
-        diff: ChainDifficulty
-    ) -> Option<(ChainDifficulty, BlockDate, BlockHash)> {
```
mzabaluev: Good, error handling.

cardano/src/block/types.rs
```diff
@@ -223,7 +223,7 @@ impl SscProof {
     }
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
```
mzabaluev: Why are these impls gone now, they were good?

storage/src/lib.rs
```diff
@@ -330,14 +330,15 @@ impl Storage {
...
     pub fn drop_loose_index_before(&mut self, diff: ChainDifficulty) {
-        if self.loose_idx.is_empty() {
-            return;
-        }
-        let drop_diff = u64::from(diff);
-        let tip_diff = u64::from(self.loose_idx[0].0);
-        let drop_after = (tip_diff - drop_diff) as usize;
-        if self.loose_idx.len() > drop_after {
-            self.loose_idx = self.loose_idx[0..drop_after].to_vec();
+        if let Some(&(idx_tip_diff, _, _)) = self.loose_idx.first() {
```
mzabaluev: I wonder if it's good to have non-descriptive three item tuples there, maybe rework them into proper structs?

cardano/src/block/types.rs
vantuz-subhuman: It just that I have added them to implement the specific code, and then code has changed, so I didn't want to introduce new stuff, without using it right away

storage/src/lib.rs
vantuz-subhuman: Makes sense, yeah 👌

vantuz-subhuman:
Discussed with @vincenthz - I don't need to use `Lookup` map at all, because I am using it only to get total number of indexed elements out, and instead I will extend and reuse the `epoch/{id}/pack` files, where I already store the list of all block offsets. Because `epoch//pack` corresponds to a specific epoch by contract, and a `Lookup` does not, so solution was hacky.

This change will remove the need for `LinkedHashMap`, so I will remove the new library dependency. But it will introduce the need for a new field that stores cached list of epoch-numbers and number of blocks in this epoch, for fast height lookup without reading all the files every time. I will combine this new field with the `loose_idx` and create something like a `chain_height_idx` struct

mzabaluev:
@vantuz-subhuman It's better if you rebase from master rather than merge.

vantuz-subhuman:
@mzabaluev, rebased into this new PR (don't like to override already existing history of review, will just close this one): https://github.com/input-output-hk/rust-cardano/pull/462

PR #456, opened 2019-02-21
Rework the network-ntt client connection as an explicit state machine

mzabaluev:
A client connection is represented by an explicit `Connection` type that implements `Future`.

Get rid of a monster generic future type and combinators holding together the asynchronous parts of a client connection. It should be easier to debug and maintain the protocol interactions in the explicitly coded state machine.

### DIFF
--- a/exe-common/src/network/ntt.rs
+++ b/exe-common/src/network/ntt.rs
@@ -30,9 +30,15 @@ impl NetworkCore {
                 // FIXME: use default executor, or take
                 // executor argument before merge.
                 let mut rt = Runtime::new().unwrap();
-                rt.spawn(connection.map(|_| {
-                    println!("Exited");
-                }));
+                rt.spawn(
+                    connection
+                        .map(|_| {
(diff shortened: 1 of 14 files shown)


qnikst:
I like the approach in general but inbound message processing will require a bit more work to be ported on the latest changes, we should use only data and take value from the requests to learn how to decode a message, protocol on its own has no required information to build Inbound::Headers and Inbound::Block.

If you want I can try to port that

exe-common/src/network/ntt.rs
```diff
@@ -30,9 +30,15 @@ impl NetworkCore {
...
-                    println!("Exited");
-                }));
+                rt.spawn(
+                    connection
+                        .map(|_| {
+                            debug!("Exited");
+                        })
+                        .map_err(|e| {
+                            error!("NTT connection error: {:?}", e);
+                        }),
```
NicolasDP: thanks for that one

PR #457, opened 2019-02-21
Mockchain delegation certificates

edolstra:
This extends the mockchain's `Transaction` type with a vector of `Certificate`s, which are signed messages for registering/deregistering a stake key, registering/retiring a stake pool, and delegating from a stake key to a stake pool. They're in `Transaction` to ensure UTxO-based protection against replays and to allow fees/deposits if needed.

Handling of these messages to update the ledger/leadership will be a subsequent PR.

### DIFF
--- a/cardano/src/block/block.rs
+++ b/cardano/src/block/block.rs
@@ -361,11 +361,11 @@ impl chain_core::property::Deserialize for Block {
 }
 
 impl chain_core::property::HasTransaction for Block {
-    type Transactions = [TxAux];
-    fn transactions(&self) -> &Self::Transactions {
+    type Transaction = TxAux;
+    fn transactions<'a>(&'a self) -> Box<Iterator<Item = &Self::Transaction> + 'a> {
         match self {
-            Block::BoundaryBlock(_) => &[],
-            Block::MainBlock(blk) => &blk.body.tx,
+            Block::BoundaryBlock(_) => Box::new([].iter()),
+            Block::MainBlock(blk) => Box::new(blk.body.tx.iter()),
         }
     }
 }
--- a/chain-core/src/property.rs
+++ b/chain-core/src/property.rs
@@ -147,13 +147,11 @@ pub trait Transaction: Serialize + Deserialize {
 /// * certificate registrations
 /// * ...
 pub trait HasTransaction {
-    /// An iterable collection of transactions provided by the block.
-    /// A reference to the `Transactions` type must be convertible to an
-    /// iterator returning references to transaction objects.
-    type Transactions: ?Sized;
+    /// The type of transactions in this block.
+    type Transaction;
 
-    /// Returns a reference that can be used to iterate over transactions in the block.
-    fn transactions(&self) -> &Self::Transactions;
+    /// Returns an iterator over the transactions in the block.
+    fn transactions<'a>(&'a self) -> Box<Iterator<Item = &Self::Transaction> + 'a>;
 }
 
 /// Updates type needs to implement this feature so we can easily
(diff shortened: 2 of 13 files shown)


edolstra:
Updated: Blocks now have a list of `Message`s, which are an enum of `SignedTransaction` and the various delegation-related certificate types.

chain-impl-mockchain/src/block.rs
```diff
@@ -13,7 +14,17 @@ pub struct Block {
     pub slot_id: BlockDate,
     pub parent_hash: Hash,
 
-    pub transactions: Vec<SignedTransaction>,
+    pub messages: Vec<Message>,
```
NicolasDP: `block_contents: Vec<Message>` I think we want to highlight that this si the content of the block and everything else is meta.

chain-impl-mockchain/src/block.rs
```diff
@@ -277,19 +286,91 @@ impl property::Deserialize for SignedBlockSummary {
...
+pub const TAG_STAKE_POOL_RETIREMENT: u8 = 6;
+
+impl property::Serialize for Message {
+    type Error = std::io::Error;
+    fn serialize<W: std::io::Write>(&self, writer: W) -> Result<(), Self::Error> {
+        use chain_core::packer::*;
+        let mut codec = Codec::from(writer);
+        match self {
+            Message::Transaction(signed) => {
+                codec.put_u8(TAG_TRANSACTION)?;
```
mzabaluev: Are tagged values better here, rather than serializing events of different types in separate (possibly type-tagged) arrays?

shell.nix
```diff
@@ -1,4 +1,4 @@
-{ nixpkgs ? fetchTarball channel:nixos-unstable
+{ nixpkgs ? fetchTarball https://github.com/NixOS/nixpkgs/archive/7fff567ee99c1f343ecdd82fef2e35fb6f50e423.tar.gz
```
mzabaluev: Ding-ding, unrelated change :)

cardano/src/block/block.rs
```diff
@@ -361,11 +361,11 @@ impl chain_core::property::Deserialize for Block {
 }
 
 impl chain_core::property::HasTransaction for Block {
-    type Transactions = [TxAux];
-    fn transactions(&self) -> &Self::Transactions {
+    type Transaction = TxAux;
+    fn transactions<'a>(&'a self) -> Box<Iterator<Item = &Self::Transaction> + 'a> {
```
mzabaluev: This introduces allocations and dynamic dispatch in iterator loops for the "production" blockchain. I'd like to avoid it if possible.

chain-core/src/property.rs
```diff
@@ -147,13 +147,11 @@ pub trait Transaction: Serialize + Deserialize {
...
-    /// A reference to the `Transactions` type must be convertible to an
-    /// iterator returning references to transaction objects.
-    type Transactions: ?Sized;
+    /// The type of transactions in this block.
+    type Transaction;
 
-    /// Returns a reference that can be used to iterate over transactions in the block.
-    fn transactions(&self) -> &Self::Transactions;
+    /// Returns an iterator over the transactions in the block.
+    fn transactions<'a>(&'a self) -> Box<Iterator<Item = &Self::Transaction> + 'a>;
```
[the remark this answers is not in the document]
edolstra: I guess that's another argument for keeping these certificates as transactions (in addition to the UTxO-based replay protection)...

chain-impl-mockchain/src/block.rs
```diff
@@ -3,6 +3,8 @@ use crate::certificate;
 use crate::key::{Hash, PrivateKey, PublicKey, Signature, Signed};
 use crate::transaction::*;
 use chain_core::property;
+use num_derive::FromPrimitive;
+use num_traits::FromPrimitive;
```
mzabaluev: Do you have to use both of these traits for this to work?

edolstra: Yeah, I think they're both required.

PR #458, opened 2019-02-21
Change net_sync to take in an Arc<RwLock<Storage>>

rooooooooob:
Previous exclusive-access &mut Storage passing was an extreme issue for
any clients of net_sync (ie hermes) that required concurrent storage
access to syncing.

As far as I can tell, only hermes uses this. I will soon have a PR to hermes using this change to fix the storage-concurrency bug there.


vantuz-subhuman:
Style is fixed in: https://github.com/input-output-hk/rust-cardano/pull/459

PR #466, opened 2019-02-25
Add to_hex and from_hex for the mockchain key.

qnikst:


NicolasDP:
@qnikst please, add rustfmt to your editor, I'd like to stop seeing commits for rust formatting.

qnikst:
applied `rustfmt` (no additional commits involved). Should I do anything about cardano::redeem::Error here?

NicolasDP:
I don't see the problem of the `redeem::Error` ...

mzabaluev:
It's OK, let's use the single domain error type unless we have a good reason to invent a parsing-specific error type.

PR #469, opened 2019-02-26
chain-core: Add method for_each_transaction to property::HasTransaction

mzabaluev:
A statically dispatched alternative to the transactions iterator.

### DIFF
--- a/chain-core/src/property.rs
+++ b/chain-core/src/property.rs
@@ -151,7 +151,16 @@ pub trait HasTransaction {
     type Transaction;
 
     /// Returns an iterator over the transactions in the block.
+    ///
+    /// Note that the iterator is dynamically allocated, and the iterator's
+    /// `next` method is invoked via dynamic dispatch. The method
+    /// `for_each_transaction` provides a statically monomorphised
+    /// alternative.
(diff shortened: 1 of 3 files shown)


chain-core/src/property.rs
```diff
@@ -151,7 +151,16 @@ pub trait HasTransaction {
     type Transaction;
 
     /// Returns an iterator over the transactions in the block.
+    ///
+    /// Note that the iterator is dynamically allocated, and the iterator's
+    /// `next` method is invoked via dynamic dispatch. The method
+    /// `for_each_transaction` provides a statically monomorphised
+    /// alternative.
     fn transactions<'a>(&'a self) -> Box<Iterator<Item = &Self::Transaction> + 'a>;
```
edolstra: Maybe we can get rid of `transactions()` now, or provide it as a generic wrapper around `for_each_transaction` outside of `HasTransactions`? That way implementations don't have to provide both `transactions` and `for_each_transaction`.

NicolasDP: that's an idea we can explore. For now we still have quite some core code utilising this API. But I think that at some point we should be able to make do with the new function, indeed.

PR #472, opened 2019-02-27
Make mockchain UTXO serializable with serde

CodeSandwich:
This will allow Jormungandr to write mockchain UTXO dump to JSON returned from REST API

### DIFF
--- a/chain-addr/src/lib.rs
+++ b/chain-addr/src/lib.rs
@@ -24,11 +24,17 @@ use std::string::ToString;
 // temporary re-use just to define
 use cardano::redeem::{self, PublicKey};
 
-use chain_core::property;
+use chain_core::property::{self, Serialize as PropertySerialize};
+
+#[cfg(feature = "generic-serialization")]
+use serde::{ser::Error as SerdeError, Serializer as SerdeSerializer};
+#[cfg(feature = "generic-serialization")]
+use serde_derive::Serialize;
 
 // Allow to differentiate between address in
 // production and testing setting, so that
 // one type of address is not used in another setting.
+#[cfg_attr(feature = "generic-serialization", derive(Serialize))]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum Discrimination {
     Production,
@@ -39,6 +45,7 @@ pub enum Discrimination {
 ///
 /// * Single address : just a single ed25519 spending public key
 /// * Group address : an ed25519 spending public key followed by a group public key used for staking
+#[cfg_attr(feature = "generic-serialization", derive(Serialize))]
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub enum Kind {
     Single(PublicKey),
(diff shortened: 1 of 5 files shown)


chain-addr/src/lib.rs
```diff
@@ -74,6 +79,7 @@ impl KindType {
 
 /// An unstructured address including the
 /// discrimination and the kind of address
+#[cfg_attr(feature = "generic-serialization", derive(Serialize))]
```
NicolasDP: Nop, this is where it gets wrong. The address is not a generic serialisable. The serde implementation is only for the `AddressReadable` type. Not for the other types.

chain-impl-mockchain/Cargo.toml
```diff
@@ -9,21 +9,21 @@ authors = [ "Nicolas Di Prima <nicolas.diprima@iohk.io>"
...
 
-[dependencies.cardano]
-path = "../cardano"
-features = [ "generic-serialization" ]
 
 [dev-dependencies]
+quickcheck = "0.8"
+rand = "0.6"
 chain-core = { path = "../chain-core", features=["property-test-api"]}
 chain-addr = { path = "../chain-addr", features=["property-test-api"]}
```
mzabaluev: I wonder if it's bad to have these duplicated as dev-dependencies with a different feature set after this change.

CodeSandwich: At the state of this commit they are needed only as dev-dependencies. Are they going to be used as normal dependencies too?

Having the same dependency in both places with different features works well, I've done this in other project.

mzabaluev: > Are they going to be used as normal dependencies too?

I don't know, it's this PR that tries to list them also as normal dependencies for some reason?

CodeSandwich: No, this PR moves them from normal deps to dev-deps

mzabaluev: Eh, I'm talking about chain-core and chain-addr.

chain-impl-mockchain/Cargo.toml
```diff
@@ -9,21 +9,21 @@ authors = [ "Nicolas Di Prima <nicolas.diprima@iohk.io>"
...
-chain-core = { path = "../chain-core" }
-chain-addr = { path = "../chain-addr" }
-rand = "0.6"
 num-traits = "0.2"
 num-derive = "0.2"
+serde = { version = "^1.0", optional = true }
+serde_derive = { version = "^1.0", optional = true }
+cardano = { path = "../cardano" }
+chain-core = { path = "../chain-core" }
+chain-addr = { path = "../chain-addr" }
```
mzabaluev: Are chain-core and chain-addr needed as regular non-dev dependencies?

CodeSandwich:
@mzabaluev The title seemed descriptive enough about changes themselves. Should I add their higher-level goal like usage in Jormungandr REST API?

mzabaluev:
> @mzabaluev The title seemed descriptive enough about changes themselves. Should I add their higher-level goal like usage in Jormungandr REST API?

Yeah, it would be generally good to understand why the change is needed.

chain-addr/src/lib.rs
CodeSandwich: I've changed `Serde` serialization of `Address` to reuse its `property::Serialize` implementation. Is it ok?

CodeSandwich:
We should wait with merge to master for Jormungandr to be able to work with master.

PR #474, opened 2019-02-28
Make mockchain UTXO serializable with serde

CodeSandwich:
This will allow Jormungandr to write mockchain UTXO dump to JSON returned from REST API

### DIFF
--- a/chain-addr/src/lib.rs
+++ b/chain-addr/src/lib.rs
@@ -24,7 +24,12 @@ use std::string::ToString;
 // temporary re-use just to define
 use cardano::redeem::{self, PublicKey};
 
-use chain_core::property;
+use chain_core::property::{self, Serialize as PropertySerialize};
+
+#[cfg(feature = "generic-serialization")]
+use serde::Serializer as SerdeSerializer;
+#[cfg(feature = "generic-serialization")]
+use serde_derive::Serialize;
 
 // Allow to differentiate between address in
 // production and testing setting, so that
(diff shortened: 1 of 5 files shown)


chain-addr/src/lib.rs
```diff
@@ -24,11 +24,17 @@ use std::string::ToString;
 // temporary re-use just to define
 use cardano::redeem::{self, PublicKey};
 
-use chain_core::property;
+use chain_core::property::{self, Serialize as PropertySerialize};
```
mzabaluev: Nitpick: I'd just use `property::Serialize` like everywhere else, the whole-module import is there anyway.

CodeSandwich: I need to import the trait so I can use its methods on structs implementing it. Without this import I would have to use UFC.

chain-addr/src/lib.rs
```diff
@@ -24,11 +24,17 @@ use std::string::ToString;
...
+
+#[cfg(feature = "generic-serialization")]
+use serde::Serializer as SerdeSerializer;
+#[cfg(feature = "generic-serialization")]
+use serde_derive::Serialize;
 
 // Allow to differentiate between address in
 // production and testing setting, so that
 // one type of address is not used in another setting.
+#[cfg_attr(feature = "generic-serialization", derive(Serialize))]
```
NicolasDP: we don't need this one

chain-addr/src/lib.rs
```diff
@@ -39,6 +45,7 @@ pub enum Discrimination {
 ///
 /// * Single address : just a single ed25519 spending public key
 /// * Group address : an ed25519 spending public key followed by a group public key used for staking
+#[cfg_attr(feature = "generic-serialization", derive(Serialize))]
```
NicolasDP: we don't need this one

PR #475, opened 2019-02-28
Basic genesis leadership and delegation handling

edolstra:
This adds a `GenesisLeaderSelection` that does the following:
    
* It keeps track of stake pool and stake key registrations and stake ("heavy-weight") delegations.
    
* At the end of every epoch, it takes a snapshot of the stake distribution per pool.
    
* Depending on the 'd' parameter, it either uses round-robin BFT rules to select a bootstrap node as slot leader, or uses genesis rules to select a stake pool as slot leader. There is no VRF/KES yet: the leader selection is completely predictable.

Other changes in this PR:

* Added a basic `Update` message for changing chain parameters (like `d`). This should eventually become an `UpdateProposal` that can be voted on, but right now the changes take effect right away and there is no security of any kind. This is mainly useful for testing the gradual migration from BFT to Genesis.

* Removed `LeaderSelection` (the wrapper over `BftLeaderSelection` and `GenesisLeaderSelection`). `GenesisLeaderSelection` implements BFT directly since it's sufficiently trivial that this is not much of an issue. Also, we have to make sure that we continue to select BFT nodes in round-robin fashion even we don't use BFT for all slots.

* Disabled diff associativity tests since random diffs don't compose properly.


cardano/src/redeem.rs
```diff
@@ -110,7 +110,7 @@ impl fmt::Display for PublicKey {
 
 pub const PRIVATEKEY_SIZE: usize = 32;
 
-#[derive(Clone)]
```
mzabaluev: I'm not sure it makes a lot of sense to define `PartialOrd`/`Ord` for binary hashes; I'd be careful here.

chain-addr/src/lib.rs
```diff
@@ -77,6 +77,15 @@ impl KindType {
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct Address(pub Discrimination, pub Kind);
```
mzabaluev: You've already implemented the expliclit fields, perhaps, and add an `Address::new` constructor?

chain-impl-mockchain/src/block.rs
```diff
@@ -13,7 +14,7 @@ pub use crate::date::{BlockDate, BlockDateParseError};
 /// with the position of that block in the chain.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Block {
-    pub slot_id: BlockDate,
+    pub slot_id: BlockDate, // FIXME: rename to 'date'
```
mzabaluev: I'd rather comment "FIXME: hide all these public members and add accessors for them". But then it's mockchain at least for the time being.

chain-impl-mockchain/src/date.rs
```diff
@@ -53,6 +61,25 @@ impl fmt::Display for BlockDate {
...
 }
 
+// FIXME: Remove this since we decided epoch durations are not in fact
+// constant.
+impl From<&BlockDate> for u64 {
+    fn from(date: &BlockDate) -> u64 {
+        date.epoch
+            .checked_mul(EPOCH_DURATION)
+            .unwrap()
+            .checked_add(date.slot_id)
```
mzabaluev: Great, checked math :1st_place_medal:

chain-impl-mockchain/src/key.rs
```diff
@@ -5,8 +5,9 @@ use cardano::hash;
 use cardano::redeem as crypto;
 use chain_core::property;
 
-/// Public key of the entity.
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+/// Public key of the entity. FIXME: do we need this wrapper around
+/// cardano::redeem::PublicKey?
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
```
mzabaluev: Same here, no real need for `Ord` traits.

chain-impl-mockchain/src/leadership/bft.rs
```diff
@@ -5,7 +5,7 @@ use crate::update::{BftSelectionDiff, ValueDiff};
 use chain_core::property::{self, LeaderSelection};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct BftRoundRobinIndex(usize);
+pub struct BftRoundRobinIndex(pub usize);
```
mzabaluev: This disclaims any invariants that may be enforced by methods on this type. Sure about exposing the inner integer?

chain-impl-mockchain/src/update.rs
```diff
@@ -243,42 +199,32 @@ mod tests {
...
-    impl Arbitrary for LeaderSelectionDiff {
-        fn arbitrary<G: Gen>(g: &mut G) -> LeaderSelectionDiff {
-            LeaderSelectionDiff {
-                bft: Arbitrary::arbitrary(g),
-            }
-        }
-    }
 
     quickcheck! {
+        /*
```
mzabaluev: Dead code, or should there be a FIXME comment stating why is it commented out?

chain-impl-mockchain/src/update.rs
```diff
@@ -243,42 +199,32 @@ mod tests {
...
         }
         fn diff_union_has_inverse_element(diff: Diff) -> bool {
             testing::update_inverse_element(diff)
         }
 
+        /*
         fn transactions_diff_union_is_associative(types: (TransactionsDiff, TransactionsDiff, TransactionsDiff)) -> bool {
             testing::update_associativity(types.0, types.1, types.2)
         }
+        */
```
mzabaluev: Same as above.

chain-impl-mockchain/src/update.rs
```diff
@@ -289,34 +235,16 @@ mod tests {
...
-        }
-        fn settings_diff_union_has_inverse_element(settings_diff: SettingsDiff) -> bool {
-            testing::update_inverse_element(settings_diff)
-        }
-
+        /*
         fn bft_selection_diff_union_is_associative(types: (BftSelectionDiff, BftSelectionDiff, BftSelectionDiff)) -> bool {
             testing::update_associativity(types.0, types.1, types.2)
         }
+        */
```
mzabaluev: As above.

chain-impl-mockchain/src/leadership/genesis.rs
```diff
@@ -0,0 +1,1481 @@
...
+    stake_keys: HashMap<PublicKey, StakeKeyInfo>,
+
+    stake_pools: HashMap<PublicKey, StakePoolInfo>,
+
+    /// The stake distribution at the end of the previous epoch.
+    stake_snapshot_n_minus_1: StakeDistribution,
+
+    /// The stake distribution at the end of the previous previous
+    /// epoch.
+    stake_snapshot_n_minus_2: StakeDistribution,
```
NicolasDP: Why not having a `stake_snapshots`: VecDeque<StakeDistribution>` ?

current stake distribution for the leadership: `[0]`. Adding a StakeDistribution is as simple as `push_back()`, consuming a `StakeDistribution` is as simple as `pop_front()`.

Can you add a separate type to manage the `StakeDistributionSnapshots` ?

chain-impl-mockchain/src/leadership/genesis.rs
```diff
@@ -0,0 +1,1481 @@
...
+#[derive(Debug)]
+pub struct GenesisLeaderSelection {
+    ledger: Arc<RwLock<Ledger>>,
+    settings: Arc<RwLock<Settings>>,
+
+    bft_leaders: Vec<PublicKey>,
+
+    stake_keys: HashMap<PublicKey, StakeKeyInfo>,
+
+    stake_pools: HashMap<PublicKey, StakePoolInfo>,
```
NicolasDP: the stake keys and the stake pools could live within a same wrapping type. Especially since their is a link between both. It would make the rest of the code easier to read too. Also, please, utilise different type for the stake_keys and stake_pools. `PublicKey` is a generic type that is being overloaded by a lot of different meaning.

chain-impl-mockchain/src/leadership/genesis.rs
```diff
@@ -0,0 +1,1481 @@
...
+    /// The stake distribution at the end of the previous previous
+    /// epoch.
+    stake_snapshot_n_minus_2: StakeDistribution,
+
+    pos: Pos,
+}
+
+// For each stake pool, the total stake value, and the value for the
+// stake pool members.
+type StakeDistribution = HashMap<PublicKey, (Value, HashMap<PublicKey, Value>)>;
```
NicolasDP: Instead of a tuple `(Value, HashMap<_, Value>)` can we have a proper type?

chain-impl-mockchain/src/key.rs
edolstra: There needed to establish an ordering between stake pool IDs (and possible stake keys in the future) for selecting a leader based on stake.

Unlikely private keys, I don't think there is a security issue in having Eq/Ord traits on them, right?

chain-impl-mockchain/src/leadership/bft.rs
edolstra: To be honest, I don't think this wrapper type is very useful to begin with...

chain-impl-mockchain/src/update.rs
edolstra: Comment added.

chain-impl-mockchain/src/leadership/genesis.rs
edolstra: Done, there now are types `StakeKeyId`, `StakePoolId`, and `LeaderId`.

chain-impl-mockchain/src/leadership/genesis.rs
edolstra: Done.

chain-impl-mockchain/src/leadership/genesis.rs
edolstra: Done, `stake_snapshots` is now a BTreeMap from epochs to stake distributions. This makes it unambiguous what epoch a snapshot refers to.

chain-impl-mockchain/src/leadership/bft.rs
edolstra: I removed the use of BftRoundRobinIndex in GenesisLeaderSelection.

chain-impl-mockchain/src/key.rs
mzabaluev: It does not feel like it's an inherent feature of the public key to be sorted that way. If leadership selection needs this weird way to sort keys, it can use a custom sorter.

I may be overthinking things here, though.

edolstra: Well, what would be a non-weird way to sort them? Lexicographical sort seems like the most straightforward thing to do.

mzabaluev: I mean that it does not make an inherent sense to sort keys in general, that's why the `PartialOrd/Ord` trait impls look dodgy.

edolstra: That's true, but OTOH, the ability to use public keys or cryptographic hashes as keys in HashMaps/BTreeMaps is pretty useful and arises quite a lot (since we often use them as identities). Given that Rust (unlike C++) unfortunately lacks the ability to specify custom comparators in containers, putting these types in Eq/Ord is the pragmatic thing to do, IMHO.

mzabaluev: I agree that it makes total sense to derive `PartialEq, Eq, Hash` for public keys, making them usable in hash tables. And perhaps you are right and `Ord` has little possibility of accidental misuse, but makes the thing qualify for B-trees.

PR #476, opened 2019-02-28
Add FromStr for Address type.

qnikst:


chain-impl-mockchain/src/key.rs
```diff
@@ -129,6 +130,13 @@ impl Hash {
     }
 }
 
+impl FromStr for Hash {
+   type Err = hash::Error;
+   fn from_str(s: &str) -> Result<Self, Self::Err> {
+      Ok(Hash(hash::Blake2b256::from_str(s)?))
+   }
+}
+
```
NicolasDP: since there is `impl std::fmt::Display for Hash` too, can you add a quickcheck property test to show that `from_str . display == id` ?

qnikst: Fixed

PR #477, opened 2019-02-28
chain-storage: Pass blocks by reference

mzabaluev:
There is no need to move `Block` values into the storage methods in our implementations, and I think this is true in general: a storage implementation would typically serialize blocks for storage, and a direct in-memory storage should be considered a toy corner case.


edolstra:
I think the original reason was that MemoryBlockStore could store the blocks directly (without serializing). However it no longer does that (and would be the exception anyway).

PR #478, opened 2019-03-01
Add immutable HAMT

vincenthz:


imhamt/src/bitmap.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    #[inline]
+    pub fn is_empty(&self) -> bool {
+        self.0 == 0
+    }
+
+    #[inline]
+    pub fn present(&self) -> usize {
+        self.0.count_ones() as usize
```
edolstra: BTW, we may want to pass `-C target-feature=+popcnt` to the compiler or add `#[target_feature(enable = "popcnt")]` to this function to get the compiler to emit code like
```
        popcntq %rdi, %rax
```
rather than
```
        movq    %rdi, %rax
        shrq    %rax
        movabsq $6148914691236517205, %rcx
        andq    %rax, %rcx
        subq    %rcx, %rdi
        movabsq $3689348814741910323, %rax
        movq    %rdi, %rcx
        andq    %rax, %rcx
        shrq    $2, %rdi
        andq    %rax, %rdi
        addq    %rcx, %rdi
        movq    %rdi, %rax
        shrq    $4, %rax
        leaq    (%rax,%rdi), %rax
        movabsq $1085102592571150095, %rcx
        andq    %rax, %rcx
        movabsq $72340172838076673, %rax
        imulq   %rcx, %rax
        shrq    $56, %rax
```

vincenthz: ha, I had no idea that it was not automatically going to popcnt (I assumed llvm would just do that). I added the attribute to the function directly

edolstra: It's probably because it requires a CPU newer than ~11 years ago.

PR #480, opened 2019-03-01
Introduce get_headers and get_blocks methods

qnikst:


network-proto/node.proto
```diff
@@ -27,6 +26,12 @@ message GetBlocksRequest {
     uint64 size = 3;
 }
 
+// Request to fetch the methods for blocks and headers.
+message DataRequest {
```
mzabaluev: I'd name this more descriptively. `BlockIds`?

network-proto/node.proto
```diff
@@ -10,13 +10,12 @@ message TipRequest {}
...
-    // Block identifier of the chain tip.
-    bytes id = 1;
-    // Block date of the chain tip.
-    string blockdate = 2;
+    // Serialized content of the tip block header.
+    bytes blockheader = 1;
 }
 
 // Request message for methods GetBlocks and GetHeaders.
+// (reserved for the future work)
```
mzabaluev: This message is perhaps not needed and should be removed.
We can restore the fields from git history when about to introduce the new random access method `GetChain`.

network-proto/node.proto
```diff
@@ -10,13 +10,12 @@ message TipRequest {}
 
 // Response message for method Tip.
 message TipResponse {
-    // Block identifier of the chain tip.
-    bytes id = 1;
-    // Block date of the chain tip.
-    string blockdate = 2;
+    // Serialized content of the tip block header.
+    bytes blockheader = 1;
```
mzabaluev: Use snake case, `block_header`?

network-core/src/server/block.rs
```diff
@@ -64,6 +64,12 @@ pub trait BlockService {
     /// accepted by this node.
     fn tip(&mut self) -> Self::TipFuture;
 
+    /// Request to load list of blocks.
+    fn pull_blocks(&mut self, ids: &[Self::BlockId]) -> Self::PullBlocksFuture;
```
mzabaluev: Please rename these to `get_blocks` and `get_headers` to be consistent with the protobuf method names.

network-proto/node.proto
mzabaluev: `blockdate` was not snake case either, but it's almost like a single word.

qnikst:
Please take another look

PR #483, opened 2019-03-01
Add get_transactions to the Tx Service.

qnikst:


network-core/src/server/transaction.rs
```diff
@@ -3,6 +3,7 @@
 use crate::codes;
 use crate::error::Code as ErrorCode;
 
+use chain_core::property;
```
mzabaluev: I tend to use `self` inside the brackets, but it's a matter of taste really.

mzabaluev: And we could just use the `property::` path consistently in this module, for clarity.

network-core/src/server/transaction.rs
```diff
@@ -12,6 +13,10 @@ use std::{error, fmt};
 /// Interface for the blockchain node service implementation responsible for
 /// validating and accepting transactions.
 pub trait TransactionService {
+
+    /// Transaction in the blockchain.
+    type Transaction: property::Transaction;
```
mzabaluev: Alternatively, import `Transaction`, it does not conflict with associated type names.

qnikst:
Thanks! fixed.

network-core/src/server/transaction.rs
mzabaluev: @qnikst TIL that you can also do `use property::Transaction as _;` go get the trait methods without importing the trait name.

NicolasDP: Only since [rust 1.33](https://blog.rust-lang.org/2019/02/28/Rust-1.33.0.html#import-as-_)

PR #484, opened 2019-03-01
fix default block_exists

NicolasDP:


edolstra:
It's a bit questionable whether `block_exists()` should return true for a block that can't actually be retrieved. However I guess this will all become academic once the genesis data becomes a real block.

NicolasDP:
Yes, vincent raised a similar note regarding this. This will change once we have the initial block.

PR #490, opened 2019-03-04
Gossip protocol implementation.

qnikst:


chain-core/src/property.rs
```diff
@@ -367,6 +367,13 @@ impl<T: Serialize> Serialize for &T {
     }
 }
 
+pub struct NodeId {
```
mzabaluev: Agree with @NicolasDP, this belongs in network-core.

chain-core/src/property.rs
```diff
@@ -367,6 +367,13 @@ impl<T: Serialize> Serialize for &T {
...
+    pub high: u64,
+    pub low: u64,
+}
+
+pub trait Gossip {}
```
mzabaluev: This is also a network thing, if it's needed at all.

network-core/src/client/gossip.rs
```diff
@@ -0,0 +1,13 @@
+use super::Error;
+
+use chain_core::property::{Gossip, NodeId};
+
+use futures::prelude::*;
+
+pub trait GossipService<T: Gossip> {
```
mzabaluev: Don't use a generic parameter here. `BlockService` was wrong about this, #486 fixes that.

network-grpc/src/client.rs
```diff
@@ -299,6 +299,15 @@ where
     T::deserialize(&mut buf).map_err(|e| core_client::Error::new(core_client::ErrorKind::Format, e))
 }
 
+fn serialize<T>(x: &T) -> Vec<u8>
```
mzabaluev: For better consistency with above, perhaps name it `serialize_to_bytes`.

chain-core/src/property.rs
NicolasDP: and add `Serialize + Deserialize` too as dependency of this trait

network-grpc/src/client.rs
NicolasDP: hence implementing the chain-core's traits : `Serialize + Deserialize` here.

qnikst:
@mzabaluev @NicolasDP Please take another look

network-core/src/server/gossip.rs
```diff
@@ -0,0 +1,70 @@
...
+
+use futures::prelude::*;
+
+use std::{error, fmt};
+
+/// Intreface for the node discovery service implementation
+/// in the p2p network.
+pub trait GossipService {
+    /// Gossip message content.
+    type Message: Gossip + Serialize + Deserialize;
```
mzabaluev: `Gossip` already implies `Serialize + Deserialize`.

network-core/src/lib.rs
```diff
@@ -5,11 +5,28 @@
...
 pub mod error;
 
 pub mod client;
 pub mod server;
 
+#[derive(Clone, Debug, PartialEq, PartialOrd)]
+pub struct NodeId(u128);
+
+impl NodeId {
+    pub fn from_bytes(high: u64, low: u64) -> Self {
```
mzabaluev: These are not bytes. There may be a `fn from_bytes(bytes: &[u8; 16])`, or more conveniently `fn from_bytes(bytes: &[u8])` with a size assertion.

I'd use the name `from_halves`.

network-grpc/src/client.rs
```diff
@@ -69,12 +72,12 @@ pub mod chain_bounds {
...
 /// make connections and perform requests towards other blockchain nodes.
-pub struct Client<B, S, E> {
+pub struct Client<B, T, S, E> {
     node: gen_client::Node<Connection<S, E, BoxBody>>,
-    _phantom: PhantomData<B>,
+    _phantom: PhantomData<(B, T)>,
 }
 
-impl<B, S, E> Client<B, S, E>
+impl<T, G, S, E> Client<T, G, S, E>
```
mzabaluev: Please use the same names for type parameters of `Client` throughout the source file. It's very confusing to have `T` represent different parameters in different places.

network-proto/node.proto
```diff
@@ -89,6 +89,18 @@ message RecordTransactionResponse {
     bytes id = 2;
 }
 
+
+// Gossip message.
+message GossipMessage {
+    message NodeId {
+      fixed64 high = 1;
+      fixed64 low = 2;
```
mzabaluev: Why do we want to specify `NodeId` in this way, and not with `bytes` of an expected size? Is there any special meaning for the 64-bit halves?

qnikst: I get this idea, but I don't think how we can preserve the current properties we want without major code complication :(

mzabaluev: What are the properties, other than the bit size of the `NodeId`?

qnikst:
I've addressed all the comments and rebased atop the latest version. @mzabaluev please take another look.

network-core/src/lib.rs
```diff
@@ -5,11 +5,37 @@
...
 pub mod error;
 
 pub mod client;
 pub mod server;
 
+#[derive(Clone, Debug, PartialEq, PartialOrd)]
+pub struct NodeId([u8; 16]);
+
+pub enum Error {
+    InvalidSize(usize),
```
mzabaluev: This `Error` is now specific to node ID errors, but it is defined on the top level.
Furthermore, there is a mod `error` which is completely unrelated.

Maybe put this stuff along with `NodeId` and `Gossip` to a submodule, tentatively named `gossip`?

network-grpc/src/client.rs
```diff
@@ -335,7 +347,26 @@ where
...
+where
+    T: Gossip,
+{
+    fn from_response(res: gen::node::GossipMessage) -> Result<(NodeId, T), core_client::Error> {
+        let node_id = match res.node_id {
+            None => panic!("no node type"),
+            Some(gen::node::gossip_message::NodeId { content }) => {
+                match NodeId::from_slice(&content) {
+                    Ok(node_id) => node_id,
+                    Err(_v) => panic!("incorrect node encoding"),
```
mzabaluev: We shall not panic from invalid network requests.
Respond with error carrying `Code::InvalidArgument` (to be added in network-core if not already).

network-core/src/lib.rs
```diff
@@ -10,6 +10,36 @@ pub mod error;
 pub mod client;
 pub mod server;
 
+pub mod gossip {
```
mzabaluev: That's more like it, though I don't like jamming stuff into `lib.rs`. Copy it out to a separate file, perhaps.
We may also want bikeshed the multiple modules named `gossip` at various levels. How about naming this module `types` and renaming `Error` to `NodeIdError`? Also try making it private, it should not be needed outside the crate.

PR #497, opened 2019-03-05
Introduce dissemination messages

qnikst:


network-core/src/server/transaction.rs
```diff
@@ -49,6 +52,9 @@ pub trait TransactionService {
         &mut self,
         ids: &[Self::TransactionId],
     ) -> Self::ProposeTransactionsFuture;
+
+    fn announce_transaction(&mut self, id: &Self::TransactionId)
```
mzabaluev: The documented design had the ability to announce multiple transactions at once.
@NicolasDP do you think it would never be useful to pack multiple transaction announcements in one protocol message? It is easily doable in gRPC by declaring the id field as repeatable.

network-proto/node.proto
```diff
@@ -26,6 +26,11 @@ message TransactionIds {
     repeated bytes id = 1;
 }
 
+// Identifier of the single transaction.
+message TransactionId {
+    bytes id = 1;
```
mzabaluev: This reminds me: in `NodeId` the field is named `content`, which is inconsistent with the other ID messages.

network-proto/node.proto
```diff
@@ -100,6 +105,8 @@ message GossipMessage {
     bytes content = 2;
 }
 
+message Confirmation {}
```
mzabaluev: In general, it would be nice to declare a separate request and response type for each gRPC call, as these serve as independent extension points for possibly adding fields later.
Streaming payload types are an exception.

network-core/src/server/transaction.rs
qnikst: Oh.. missed that, I can send multiple transactions, that can be even easier for me

qnikst:
@mzabaluev I've fixed everything, please take another look.

network-proto/node.proto
```diff
@@ -100,6 +100,9 @@ message GossipMessage {
     bytes content = 2;
 }
 
+message ConfirmAnnounceBlock {}
+message ConfirmAnnounceTx {}
```
mzabaluev: I'd just keep to the pattern elsewhere (picked up from gRPC guides) and name them
`AnnounceBlockResponse` and `AnnounceTransactionResponse`.

network-proto/node.proto
```diff
@@ -121,4 +124,6 @@ service Node {
     rpc RecordTransaction (RecordTransactionRequest) returns (RecordTransactionResponse);
```
mzabaluev: While we are at it, we can remove `ProposeTransactions` and `RecordTransaction` as these are replaced by the new messages.

network-proto/node.proto
```diff
@@ -121,4 +124,6 @@ service Node {
...
 
     rpc Gossip (GossipMessage) returns (GossipMessage);
+    rpc AnnounceBlock (Header) returns (ConfirmAnnounceBlock);
+    rpc AnnounceTransaction (TransactionIds) returns (ConfirmAnnounceTx);
```
mzabaluev: Name `AnnounceTransactions` then?

PR #499, opened 2019-03-05
Update the block serialisation to match the specifications

NicolasDP:


chain-core/src/packer.rs
```diff
@@ -3,79 +3,184 @@
...
 pub struct Codec<I>(I);
 impl<I> Codec<I> {
     pub fn into_inner(self) -> I {
         self.0
     }
 }
+
+pub struct Buffered<I: std::io::Write>(I, Codec<Vec<u8>>);
+
+pub struct Hole<T> {
```
mzabaluev: Would be nice to add a doc comment explaining the purpose of the type.

chain-impl-mockchain/src/block/header.rs
```diff
@@ -0,0 +1,300 @@
...
+pub type BlockContentHash = Hash;
+pub type BlockId = Hash;
+pub type BlockContentSize = u32;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct BlockVersion(u16);
+
+pub const BLOCK_VERSION_CONSENSUS_NONE: BlockVersion = BlockVersion::new(0x0000_0000);
+pub const BLOCK_VERSION_CONSENSUS_BFT: BlockVersion = BlockVersion::new(0x0000_0001);
+pub const BLOCK_VERSION_CONSENSUS_GENESIS_PRAOS: BlockVersion = BlockVersion::new(0x0000_0002);
```
mzabaluev: Why not make this an enum with numeric variants?

chain-impl-mockchain/src/block/mod.rs
```diff
@@ -1,12 +1,18 @@
...
-use num_derive::FromPrimitive;
-use num_traits::FromPrimitive;
+
+mod header;
+mod message;
+
+pub use self::header::{
+    BftProof, BlockContentHash, BlockContentSize, BlockId, BlockVersion, Common, GenesisPraosProof,
+    Header, Proof, BLOCK_VERSION_CONSENSUS_BFT, BLOCK_VERSION_CONSENSUS_GENESIS_PRAOS,
+    BLOCK_VERSION_CONSENSUS_NONE,
```
mzabaluev: Gah, enums are just better for this.

PR #504, opened 2019-03-05
Chain crypto step2

vincenthz:
still missing instance for PublicKeyError -> Error


mzabaluev:
Please look into the test failures, they seem to be introduced by recent `chain_crypto` changes.

PR #506, opened 2019-03-06
Changes towards support of the gossip service.

qnikst:
Rename methods and introduce generic functions for
convertion gossip id's and types.

### DIFF
--- a/network-core/src/client/gossip.rs
+++ b/network-core/src/client/gossip.rs
@@ -1,6 +1,6 @@
 use super::Error;
 
-use super::super::gossip::{Gossip, NodeId};
+use super::super::gossip::{self, Gossip};
 
 use futures::prelude::*;
 
@@ -8,8 +8,8 @@ pub trait GossipService {
     type Gossip: Gossip;
 
     /// Future that represent an asynchronous gossip request.
-    type GossipFuture: Future<Item = (NodeId, Self::Gossip), Error = Error>;
+    type GossipFuture: Future<Item = (gossip::NodeId, Self::Gossip), Error = Error>;
 
     /// Request to announce our own gossip and bring back result.
-    fn gossip(&mut self, node_id: &NodeId, gossip: &Self::Gossip) -> Self::GossipFuture;
+    fn gossip(&mut self, node_id: &gossip::NodeId, gossip: &Self::Gossip) -> Self::GossipFuture;
 }
--- a/network-core/src/gossip.rs
+++ b/network-core/src/gossip.rs
@@ -22,4 +22,13 @@ impl NodeId {
     }
 }
 
-pub trait Gossip: Serialize + Deserialize {}
+pub trait Gossip: Serialize + Deserialize {
+    /// Type that represents NodeId in the gossip message.
+    type NodeId: Sized;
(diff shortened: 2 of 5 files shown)


network-core/src/client/gossip.rs
```diff
@@ -1,15 +1,15 @@
...
+use super::super::gossip::{self, Gossip};
 
 use futures::prelude::*;
 
 pub trait GossipService {
     type Gossip: Gossip;
 
     /// Future that represent an asynchronous gossip request.
-    type GossipFuture: Future<Item = (NodeId, Self::Gossip), Error = Error>;
+    type GossipFuture: Future<Item = (gossip::Id, Self::Gossip), Error = Error>;
```
mzabaluev: I don't like the name `Id`, it's too short to mean anything. Could be confused with the ID of the gossip message by a casual reader.

network-core/src/gossip.rs
```diff
@@ -1,25 +1,33 @@
...
         self.0[..].into()
     }
 }
 
-pub trait Gossip: Serialize + Deserialize {}
+pub trait Gossip: Serialize + Deserialize {
+    type Id: Sized;
+    type Node;
+
+    fn to_node_id(id: &Self::Id) -> Id;
```
mzabaluev: So it's `node_id` in methods, perhaps `NodeId` was the consistent type name.

mzabaluev: Why do we need this duality of id types? Not possible to live with just the associated type on `Gossip` as the node ID?

network-core/src/gossip.rs
```diff
@@ -1,25 +1,33 @@
...
 
     pub fn to_bytes(&self) -> Vec<u8> {
         self.0[..].into()
     }
 }
 
-pub trait Gossip: Serialize + Deserialize {}
+pub trait Gossip: Serialize + Deserialize {
+    type Id: Sized;
+    type Node;
```
mzabaluev: No documentation, gives no idea of the purpose of `Node`.
"Gossip information about a node"?

network-core/src/gossip.rs
```diff
@@ -1,25 +1,33 @@
...
+
+    fn to_node_id(id: &Self::Id) -> Id;
+    fn from_node_id(id: &Id) -> Self::Id;
+
+    fn from_nodes(iter: Box<Iterator<Item = Self::Node>>) -> Self;
```
mzabaluev: This could be generic rather than forcing an allocation:

```rust
    fn from_nodes<I>(iter: I) -> Self
    where I: IntoIterator<Item = Self::Node>
    {
    }
```

network-core/src/gossip.rs
```diff
@@ -1,7 +1,7 @@
 use chain_core::property::{Deserialize, Serialize};
 
 #[derive(Clone, Debug, PartialEq, PartialOrd)]
-pub struct NodeId([u8; 16]);
+pub struct NodeId(pub [u8; 16]);
```
mzabaluev: I'd rather avoid promiscuous mutability here if we can help it.

mzabaluev:
Needs rustfmt.

PR #508, opened 2019-03-07
add way to generically manage the block version status from the settings and block

NicolasDP:


cardano/src/block/block.rs
```diff
@@ -94,6 +95,13 @@ impl chain_core::property::Header for BlockHeader {
             BlockHeader::MainBlockHeader(ref header) => header.consensus.slot_id.into(),
         }
     }
+
+    fn version(&self) -> Self::Version {
+        match self {
+            BlockHeader::BoundaryBlockHeader(ref header) => unimplemented!(),
```
mzabaluev: This will crash on any boundary block?

mzabaluev: Think of returning `Option<Self::Version>` if Cardano boundary blocks can't have version by design.

chain-core/src/property.rs
```diff
@@ -60,11 +60,17 @@ pub trait Header: Serialize + Deserialize {
     /// The block date.
     type Date: BlockDate;
 
+    /// the type associated to the version of a block
+    type Version;
```
mzabaluev: Might want to add an empty tag trait `Version` and bind this associated type with it, just like we do for `Id` and `Date`.

chain-impl-mockchain/src/leadership/mod.rs
```diff
@@ -49,6 +51,12 @@ impl property::Deserialize for LeaderId {
     }
 }
 
+impl AsRef<PublicKey<Ed25519Extended>> for LeaderId {
```
mzabaluev: This seems unrelated.

PR #509, opened 2019-03-07
Fix ledger tests

edolstra:


chain-impl-mockchain/src/ledger.rs
```diff
@@ -137,113 +138,113 @@ pub mod test {
...
+        };
+        assert_eq!(
+            Err(Error::InvalidSignature(utxo0, output0, witness)),
+            ledger.diff_transaction(&signed_tx)
+        )
+    }
 
-    */
+    #[test]
+    fn cant_loose_money() {
```
mzabaluev: Typo: lose. But you can't loose that money, either :stuck_out_tongue_winking_eye:

PR #513, opened 2019-03-07
Add account ledger

vincenthz:


chain-impl-mockchain/src/account.rs
```diff
@@ -0,0 +1,181 @@
...
+
+/// Account Identifier (also used as Public Key)
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Identifier(PublicKey<AccountAlg>);
+
+/// Account Secret Key
+pub struct Secret(SecretKey<AccountAlg>);
+
+#[derive(Clone)]
+pub struct State {
```
edolstra: `State` is a bit vague, maybe `AccountState` would be better (or even just `Account`)?

chain-impl-mockchain/src/account.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    /// Add a new account into this ledger.
+    ///
+    /// If the identifier is already present, error out.
+    pub fn add_account(
+        &self,
+        account: &Identifier,
+        initial_value: Value,
+    ) -> Result<Self, LedgerError> {
+        self.0
```
edolstra: Can `initial_value` be zero?

chain-impl-mockchain/src/account.rs
vincenthz: it was `AccountState` originally, but I stripped the `Account` prefix in favor of using the module system

PR #514, opened 2019-03-08
[IOHKCRB-5] Documentation for header functions

enzoc4:
---
name: [IOHKCRB-5] Documentation for the C side of things
---
# Description
This PR introduces C api documentation in the cardano.h so developers could easily understand what they are calling.

# Proposal
Add documentation in the C header file that can be used to generate a HTML document with Doxygen, and add documentation in places where it's missing.

# Mode of use
The HTML documentation is generated in the cardano-c/docs/html directory with the `doxygen` command in the **cardano-c** directory.

# Examples

doxygen

## Comparison

doxy_account_create

doxy_account_create_rust


cardano-c/Doxyfile
```diff
@@ -0,0 +1,2427 @@
+# Doxyfile 1.8.11
```
NicolasDP: can you regenerate the Doxyfile without the doc ?

enzoc4: do you mean without the auto generated comments for every variable?

PR #515, opened 2019-03-10
Add bech32 prefixes to crypto keys 

CodeSandwich:
This is needed for exporting and importing bech32-encoded keys. Currently only utilized during generation of private and public keys in CLI, may have more uses.

### DIFF
--- a/chain-crypto/src/algorithms/ed25519.rs
+++ b/chain-crypto/src/algorithms/ed25519.rs
@@ -1,7 +1,7 @@
 use crate::key::{AsymmetricKey, PublicKeyError, SecretKeyError};
 use crate::sign::{SignatureError, SigningAlgorithm, Verification, VerificationAlgorithm};
 use cryptoxide::ed25519;
-use rand_core::{CryptoRng, RngCore};
+use rand::{CryptoRng, RngCore};
 
 use ed25519_bip32::XPub;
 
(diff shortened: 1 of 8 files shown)


chain-crypto/src/algorithms/ed25519.rs
```diff
@@ -47,6 +47,9 @@ impl AsymmetricKey for Ed25519 {
     type Secret = Priv;
     type Public = Pub;
 
+    const SECRET_BECH32_HRP: &'static str = "ed25519_secret";
+    const PUBLIC_BECH32_HRP: &'static str = "ed25519_public";
```
mzabaluev: I don't see any of these new private constants used anywhere?

CodeSandwich: They are not private, traits always define only public things. They're already used in jormungandr branch with key generation in CLI.

mzabaluev: Ah, sorry, these are associated const items.

chain-crypto/src/key.rs
```diff
@@ -19,6 +19,9 @@ pub trait AsymmetricKey {
     type Secret: AsRef<[u8]> + Clone;
     type Public: AsRef<[u8]> + Clone + PartialEq + Eq + Hash;
 
+    const SECRET_BECH32_HRP: &'static str;
+    const PUBLIC_BECH32_HRP: &'static str;
```
NicolasDP: what HRP stands for ?

CodeSandwich: Human readable part. This name and its abbreviation seem standard in bech32 nomenclature.

NicolasDP: thanks

PR #516, opened 2019-03-11
Add Clone and Debug to NodeIdError

qnikst:


NicolasDP:
since you are at that. can you also add `Display` and `Error` traits impl for this type?

qnikst:
I've added `Error` and `Display` traits.

PR #518, opened 2019-03-12
Make BlockStore usable as a trait object

edolstra:
This requires getting rid of the generic type parameter in `get_path_to_nth_ancestor()`, dropping the `Self` return type in `iterate_range()` and dropping `Sized`.

### DIFF
--- a/chain-storage/src/store.rs
+++ b/chain-storage/src/store.rs
@@ -36,7 +36,7 @@ pub struct BackLink<Id: BlockId> {
     pub block_hash: Id,
 }
 
-pub trait BlockStore: std::marker::Sized {
+pub trait BlockStore {
     type Block: Block;
 
     /// Write a block to the store. The parent of the block must exist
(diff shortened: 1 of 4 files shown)


chain-storage/src/store.rs
```diff
@@ -222,42 +222,39 @@ pub trait BlockStore: std::marker::Sized {
...
                     to_depth: to_info.depth,
                     cur_depth: to_info.depth - distance,
                     pending_infos: vec![to_info],
                 })
             }
         }
     }
+
+    // See https://stackoverflow.com/questions/42121299/provided-method-casting-self-to-trait-object
+    fn as_trait(&self) -> &BlockStore<Block = Self::Block>;
```
mzabaluev: Note that there were a few other solutions that don't require a bogus method. I'd go with a helper blanket-implemented trait.

NicolasDP: you meant something like:

```rust
pub trait BlockStore {
  fn as_trait(&self) -> &BlockStore<Block = Self::Block> { self }
}
```

would that work ?

edolstra: That doesn't work since `self` is not `Sized`, and adding a `Sized` bound to `as_trait` makes it unavailable in trait objects.

PR #520, opened 2019-03-13
Use popcount only on x86.

qnikst:


edolstra:
Can you apply rustfmt? Travis is complaining about it.

qnikst:
Fixed

PR #522, opened 2019-03-15
Initial version of the transaction builder.

qnikst:
Current version of the transaction builder is designed to be
as simple as possible and contain only barebone API. So it can
be later improved to fit all the needs.
Builder only works with stable fee algorithms i.e. the ones that
can perfectly estimate the fee cost. The only provided fee
algorithm exactly matches that criterion. It sums up inputs
and outputs count and calculates the cost.

### DIFF
--- a/chain-impl-mockchain/src/fee.rs
+++ b/chain-impl-mockchain/src/fee.rs
@@ -1,0 +1,35 @@
+use crate::transaction as tx;
+use crate::value::Value;
+use chain_addr::Address;
+
+/// Linear fee using the basic affine formula
+/// `COEFFICIENT * (COUNT(tx inputs) + COUNT(tx outputs)) + CONSTANT`.
+#[derive(PartialEq, Eq, PartialOrd, Debug, Clone, Copy)]
+pub struct LinearFee {
+    pub constant: u64,
+    pub coefficient: u64,
+}
+
+impl LinearFee {
+    pub fn new(constant: u64, coefficient: u64) -> Self {
+        LinearFee {
+            constant,
+            coefficient,
+        }
+    }
+}
+
+pub trait FeeAlgorithm {
+    fn calculate_for(&self, tx: &tx::Transaction<Address>) -> Option<Value>;
+}
+
+impl FeeAlgorithm for LinearFee {
+    fn calculate_for(&self, tx: &tx::Transaction<Address>) -> Option<Value> {
+        let msz = (tx.inputs.len() as u64).checked_add(tx.outputs.len() as u64)?;
+        let fee = self
+            .coefficient
+            .checked_mul(msz)?
+            .checked_add(self.constant)?;
+        Some(Value(fee))
+    }
+}
--- a/chain-impl-mockchain/src/txbuilder.rs
+++ b/chain-impl-mockchain/src/txbuilder.rs
@@ -1,0 +1,209 @@
+use crate::fee::FeeAlgorithm;
+use crate::key::SpendingSecretKey;
+use crate::transaction as tx;
+use crate::value::{Value, ValueError};
+use chain_addr::Address;
+use std::{error, fmt};
+
+/// Possible error for the builder.
+#[derive(Debug, Clone)]
+pub enum Error {
+    TxInvalidNoInput,
+    TxInvalidNoOutput,
+    TxNotEnoughTotalInput,
+    MathErr(ValueError),
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Error::TxInvalidNoInput => write!(f, "transaction has no inputs"),
+            Error::TxInvalidNoOutput => write!(f, "transaction has no outputs"),
+            Error::TxNotEnoughTotalInput => write!(f, "not enough input for making transaction"),
+            Error::MathErr(v) => write!(f, "error in arithmetics {:?}", v),
+        }
+    }
+}
+
+impl error::Error for Error {
+    fn source(&self) -> Option<&(dyn error::Error + 'static)> {
+        None
+    }
+}
+
+/// Output policy to be used in transaction. This policy is used then
+/// there is positive balance on in the OUTPUT+FEE-INPUT. Policy
+/// explains how to use that balance. Rember that policy application
+/// may change the amount of the fee.
+#[derive(Debug, Clone)]
+pub enum OutputPolicy {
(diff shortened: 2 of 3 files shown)


chain-impl-mockchain/src/txbuilder.rs
```diff
@@ -0,0 +1,192 @@
...
+                    value: Value(0),
+                });
+                let fee = fee_algorithm.calculate_for(&tx);
+                match balance(&tx, fee) {
+                    Ok(Balance::Positive(value)) => {
+                        self.0.outputs.push(tx::Output { address, value });
+                        return Ok((Balance::Zero, TransactionFinalizer::new(self.0)));
+                    }
+                    _ => {
+                        return Ok((Balance::Positive(pos), TransactionFinalizer::new(tx)));
```
CodeSandwich: When balance with 0-value TX turns negative, do we still want to perform the 0-value TX? It will get rejected. This strategy seems fine when the result is `Balance::Zero`. I'm not sure what to do if balance is negative. Remove 0-value TX and add the excessive value to a random output? Assign excess to output address 0 and send it to junkyard forever littering UTXO space?

chain-impl-mockchain/src/txbuilder.rs
```diff
@@ -0,0 +1,192 @@
...
+use crate::key::SpendingSecretKey;
+use crate::transaction as tx;
+use crate::value::{Value, ValueError};
+use chain_addr::Address;
+
+/// Possible error for the builder.
+#[derive(Debug, Clone)]
+pub enum Error {
+    TxInvalidNoInput,
+    TxInvalidNoOutput,
```
CodeSandwich: These 2 errors are not used anywhere, I hope that's ok? Should they be cleaned up?

chain-impl-mockchain/src/fee.rs
```diff
@@ -0,0 +1,68 @@
...
+            coefficient,
+        }
+    }
+}
+
+#[derive(PartialEq, Eq, PartialOrd, Debug, Clone, Copy)]
+pub struct Milli(u64);
+impl Milli {
+    pub fn new(i: u64, f: u64) -> Self {
+        Milli(i * 1000 + f % 1000)
```
CodeSandwich: Do we need to silently overflow the fraction part with `% 1000`? Maybe we could `i * 1000 + f`?

chain-impl-mockchain/src/fee.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub fn new(constant: Milli, coefficient: Milli) -> Self {
+        LinearFee {
+            constant,
+            coefficient,
+        }
+    }
+}
+
+#[derive(PartialEq, Eq, PartialOrd, Debug, Clone, Copy)]
+pub struct Milli(u64);
```
CodeSandwich: Should we worry about overflows? Here we put `Value`, which is u64 multiplied by 1000 and store it as `u64`, which could cause overflow. Maybe we should error-guard all the operations like in `Value` itself?

CodeSandwich: How about storing internal value as u128?

chain-impl-mockchain/src/txbuilder.rs
qnikst: The situation here, is that few lines above we have already generated transaction with the positive value, so we can always fall back to that transaction.

So I had the following consideration: creating empty transaction doesn't benefit the sender at all so in `Zero` case we fall back to the previous approach, where we had a bit of positive balance. This can potentially be interesting to the node performing transaction.

In general, if the fee calculation procedure is more complex we will need to perform clever branch analysis.

chain-impl-mockchain/src/txbuilder.rs
qnikst: will use them now :) thanks

chain-impl-mockchain/src/fee.rs
qnikst: Not sure I just took millis from the `cardano` crate.

chain-impl-mockchain/src/txbuilder.rs
CodeSandwich: But as far as I can see in this specific case (line 125) we return `tx`, which has zero value output added (line 114), we don't fall back to the slightly positive transaction `self.0` without zero value output.

qnikst: Thanks, it should be `self.0` here!

CodeSandwich: What happens with the excess if we create transaction with positive balance? Does it get lost forever? Are zero-value outputs ultimate dust or they get discarded before reaching UTXO and therefore safe to generate?

qnikst: As @NicolasDP it's perfectly ok and we can loose money, also as far as I know that can be used by the node applying a transaction so excess of the money may be  used to increase transaction priority as nodes may be more interested to commit that one.

chain-impl-mockchain/src/fee.rs
vincenthz: the approach is kind of tied to the fact that cardano's coin are limited at 45B .. here we don't have this luxury, so simple solution is just throwing away the fractional part.

CodeSandwich: @vincenthz I don't quite understand your comment. Why having no cap on Value makes throwing away fractional part a good idea?

qnikst:
@CodeSandwich please take another look.

qnikst:
@CodeSandwich I've addressed all the comments, removed millis added overflow check and updated code accordingly, please take another look.

chain-impl-mockchain/src/txbuilder.rs
```diff
@@ -0,0 +1,174 @@
...
+        TransactionFinalizer(tx::SignedTransaction {
+            transaction,
+            witnesses: vec![],
+        })
+    }
+
+    /// Sign transaction.
+    pub fn sign(&mut self, pk: &SpendingSecretKey) {
+        // TODO: check if signature is required.
+        // TODO: check if signature matches address.
```
NicolasDP: this will be proven difficult here as you don't have enough information to check this. The input does not contain the address (so you don't have the public key associated here). Might be worth thinking of other options for future improvements, something like:

```rust
pub fn sign<F>(&mut self, sign: F)
where F: FnMut(&Input) -> Witness
{
    // ...
}
```

qnikst:
@NicolasDP I've addressed all your comments, please take another look.

PR #523, opened 2019-03-15
[IOHKCRB-7] Add test of transaction creation

enzoc4:
---
name: [IOHKCRB-7] Add test of transaction creation
---

# Description
## Add 3 tests to transaction building functionality
To the return status of the builder functions that return `cardano_result`, to assert that they report errors correctly.

**Note**: As there is no way to check the balance of the builder it is not possible to make assertions over the effects of the functions that add inputs and outputs. 

### Problem
Currently, there are a few functions that allocate data structures that cannot be deallocated.
- `cardano_account_generate_addresses`
- `cardano_transaction_builder_finalize`
- `cardano_transaction_finalized_new`

### Solution
- `cardano_account_delete_addresses`
- `cardano_transaction_delete`
- `cardano_transaction_finalized_delete`

## Add valgrind to the test runner (_script.sh_)

To be able to check for memory leaks in the tests and detect possible deallocation issues, set the environment variable VALGRIND to true to use it.
### Example:
`VALGRIND=true ./cardano-c/test.sh`

### DIFF
--- a/cardano-c/cardano.h
+++ b/cardano-c/cardano.h
@@ -1,7 +1,7 @@
 /*! \file cardano.h
 */
 #ifndef CARDANO_RUST_H
-# define CARDANO_RUST_H
+#define CARDANO_RUST_H
 /* Basic Types */
 
 #ifdef __cplusplus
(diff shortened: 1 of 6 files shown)


NicolasDP:
Just apply rustfmt to this PR so we can merge it in

enzoc4:
> Just apply rustfmt to this PR so we can merge it in

I think the test was failing because of some issue in master at the time I originally pushed. I did a new push and it is passing now.

NicolasDP:
@enzoc4 if you can resolve the conflict (I merged #514  and it created this conflict).

enzoc4:
I merged master and resolved the conflict, but now the build is breaking. I think the problem is in the commits that I merged in?

PR #532, opened 2019-03-19
Implement certificates support in transaction builder

qnikst:


chain-impl-mockchain/src/fee.rs
```diff
@@ -0,0 +1,84 @@
+use crate::certificate::Certificate;
+use crate::transaction as tx;
+use crate::value::Value;
+use chain_addr::Address;
+use std::ops::{Add, Mul};
+
+/// Linear fee using the basic affine formula
+/// `A * bytes(COUNT(tx.inputs) + COUNT(tx.outputs)) + CONSTANT + C*COUNT(certificates)`.
```
NicolasDP: here it says `bytes(...)`. but in the implementation at the bottom of this file it is the `A * (num_inputs + num_outputs)`. Update the doc please.

qnikst: Fixed!

qnikst: Actuall fixed in a tx-build PR, I'll rebase this one soonish (after a finishing adding proper serialization of the certificate).

qnikst:
PR is ready for the review now.

PR #537, opened 2019-03-19
Bidirectional subscription

mzabaluev:
Added bidirectional subscription methods to gRPC protobuf definitions
and network-core API.
The bidirectional subscriptions replace the `Announce*` methods and the unidirectional `subscribe_to_blocks` method.


mzabaluev:
The block and transaction subscription support is ready. Changes for node gossip need more redesign, so I'm splitting them off to a separate PR.

PR #542, opened 2019-03-20
[IOHKCRB-12] Improve error handling for wallet creation

enzoc4:
---
name: [IOHKCRB-12] Add test of transaction creation
---

# Description

## Error handling

- Make `cardano_wallet_new` return the created wallet as an out parameter and use the return value for the error code. This way it no longer returns a null pointer if the size of the entropy array is not valid.
- Add the tests to validate that the error code is consistent

### DIFF
--- a/cardano-c/src/wallet.rs
+++ b/cardano-c/src/wallet.rs
@@ -9,7 +9,7 @@ use std::os::raw::c_char;
 use std::{ffi, ptr, slice};
 
 use address::ffi_address_to_base58;
-use types::{AccountPtr, WalletPtr};
+use types::{AccountPtr, CardanoResult, WalletPtr};
 
 /* ******************************************************************************* *
  *                                  Wallet object                                  *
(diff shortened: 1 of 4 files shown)


NicolasDP:
Looking good, you might need to rebase your branch. The previous merge (#523 ) seems to have conflict with this one.

cardano-c/src/wallet.rs
```diff
@@ -25,27 +25,30 @@ use types::{AccountPtr, WalletPtr};
...
-        Err(_) => return ptr::null_mut(),
+        Err(_) => return CardanoResult::failure(),
         Ok(e) => e,
     };
 
     let wallet = bip44::Wallet::from_entropy(&entropy, &password, hdwallet::DerivationScheme::V2);
 
     let wallet_box = Box::new(wallet);
-    Box::into_raw(wallet_box)
+    unsafe { ptr::write(wallet_out, Box::into_raw(wallet_box)) };
```
vincenthz: might be a good idea to check the wallet_out pointer being non NULL and returning a friendly CardanoResult error if that's the case ?

vincenthz:
Just need rebasing and then we can merge

cardano-c/src/wallet.rs
enzoc4: I see... The behavior is undefined in that case, wouldn't a panic with a message be better though? I think a null pointer here is most probably a bug in the caller?
Should I always check for null pointers?

PR #550, opened 2019-03-21
[IOHKCRB-16] Unimplemented functions in XPrv module

enzoc4:
---
name: [IOHKCRB-16] Unimplemented functions in XPrv module
---
# Description

- Fix compilation issues in *key.rs* and export it.
- Implemented `cardano_xprv_to_bytes`.
- Associated function `cardano_xprv_bytes_delete` to free the allocated memory.
- Return `cardano_error` in `cardano_xprv_from_bytes` to contemplate failure. <sup>1</sup>

<sup>1</sup> _The error can happen because I used `from_bytes_verified` (instead of `from_bytes`), previously it was using `from_slice` but I didn't compile because it is  a private method._

### DIFF
--- a/cardano-c/src/key.rs
+++ b/cardano-c/src/key.rs
@@ -1,10 +1,10 @@
 use cardano::hdwallet;
 use std::{
     ffi,
-    ptr,
     os::raw::{c_char, c_int},
+    ptr, slice,
 };
-use types::{XPrvPtr, XPubPtr};
+use types::{CardanoResult, XPrvPtr, XPubPtr};
 
 #[no_mangle]
 pub extern "C" fn cardano_xprv_derive(c_xprv: XPrvPtr, index: u32) -> XPrvPtr {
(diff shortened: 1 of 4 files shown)


cardano-c/src/key.rs
```diff
@@ -15,22 +15,47 @@ pub extern "C" fn cardano_xprv_derive(c_xprv: XPrvPtr, index: u32) -> XPrvPtr {
...
+
+    //Avoid running the destructor
+    std::mem::forget(vector);
+    ptr
+}
+
+#[no_mangle]
+pub extern "C" fn cardano_xprv_bytes_delete(bytes: *mut u8) {
+    let vector = unsafe { Vec::from_raw_parts(bytes, hdwallet::XPRV_SIZE, hdwallet::XPRV_SIZE) };
+    std::mem::drop(vector)
```
NicolasDP: Would be nice to zero the value here. We have a function for this in [`cardano::util::securemem::zero`](https://github.com/input-output-hk/rust-cardano/blob/master/cardano/src/util/securemem.rs#L4) that will do the work.

Simply add:

```rust
cardano::util::securemem::zero(&mut vector);
```

enzoc4: Done!

cardano-c/src/key.rs
```diff
@@ -15,22 +15,49 @@ pub extern "C" fn cardano_xprv_derive(c_xprv: XPrvPtr, index: u32) -> XPrvPtr {
...
     let xprv_data = unsafe { slice::from_raw_parts(c_xprv, hdwallet::XPRV_SIZE) };
-    match hdwallet::XPrv::from_slice(xprv_data) {
+    let array = {
+        let mut array = [0u8; 96];
+        array.copy_from_slice(xprv_data);
+        array
+    };
+    match hdwallet::XPrv::from_bytes_verified(array) {
         Ok(r) => {
+            println!("ok");
```
NicolasDP: you might want to get rid of this debug message.

(btw, there is a `dbg!` macro you can use that are armless in release mode -- unlike a `println`)

vincenthz:
I will remove the debug print

PR #551, opened 2019-03-22
Rework the node gossip protocol

mzabaluev:
Make node gossip propagated bidirectionally, like blocks and
transactions.
Each gossip message carries information on a single node.

### DIFF
--- a/network-grpc/proto/node.proto
+++ b/network-grpc/proto/node.proto
@@ -50,14 +50,12 @@ message Message {
     bytes content = 1;
 }
 
-// Gossip message with information of a single network node.
-message NodeGossip {
-    // The node identifier.
-    bytes id = 1;
-    // The IP address.
-    string host = 2;
-    // Port number.
-    uint32 port = 3;
+// Gossip message with information on nodes in the network.
+message Gossip {
+    // ID of the sender node (TODO: replace with gRPC metadata).
+    bytes sender = 1;
+    // Serialized descriptions of nodes.
+    repeated bytes nodes = 2;
 }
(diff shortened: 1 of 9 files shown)


network-grpc/proto/node.proto
```diff
@@ -49,14 +49,14 @@ message Transaction {
...
-    bytes content = 2;
+// Gossip message with information of a single network node.
+message NodeGossip {
+    // The node identifier.
+    bytes id = 1;
+    // The IP address.
+    string host = 2;
+    // Port number.
+    uint32 port = 3;
 }
```
NicolasDP: so actually here we want as before. A gossip message is:

* an ID; (the ID of the sender);
* a list of tuple (Id, Node) (the gossips);

The Sender always include itself in the list. This is how we get the information on the Node that contacted us.

mzabaluev: Is `Node` going to be an opaque blob, or should we detail some fields in protobuf? It can be specified as a multiaddr string.

NicolasDP: it is an opaque block. https://docs.rs/poldercast/0.2.1/poldercast/struct.Node.html
They implement `serde`. Just utilise `bincode` to encode/decode the whole `btreemap` and the `id`.

mzabaluev: It's back to sender ID + opaque nodes now. I did not segregate the ID field out of the node payloads as poldercast `Node` would serialize as a binary blob containing the id; no need to duplicate that information in protobuf.

PR #553, opened 2019-03-24
remove debug noise

NicolasDP:


vincenthz:
already in ledger-unification

PR #554, opened 2019-03-25
network: Disambiguate subscription methods

mzabaluev:
Prefix names of subscription methods in network-core APIs with the object of subscription.
This helps to disambiguate methods on the client side. The server side API
is also changed for consistency.


NicolasDP:
@mzabaluev don't forget to use rustfmt

PR #575, opened 2019-03-26
[IOHKCRB-15] transaction observers

enzoc4:
# Add functions to observe the state of a txbuilder
- `cardano_transaction_builder_balance`
- `cardano_transaction_builder_balance_without_fees`
- `cardano_transaction_builder_get_input_total`
- `cardano_transaction_builder_get_output_total`

Mapping to the following functions in the txbuilder module: 
- `pub fn get_input_total(&self) -> Result<Coin>`
- `pub fn get_output_total(&self) -> Result<Coin>`
- `pub fn balance<'a, F: FeeAlgorithm>(&self, f: &'a F) -> Result<CoinDiff>`
- `pub fn balance_without_fees(&self) -> Result<CoinDiff>`

The balance is represented as the following C non-opaque struct, where the sign is used to represent the `Positive` (1), `Negative` (-1) and `Zero` (0) variants of the possible results (represented by a `CoinDiff`).
```C
typedef struct cardano_transaction_balance {
    int32_t sign;
    uint64_t value;
} cardano_transaction_balance_t
```
Altough the struct is non-opaque, it is allocated in the heap by the Rust code, to be consistent with the existing allocation, and the memory is freed with 
`cardano_transaction_balance_delete`

### DIFF
--- a/cardano-c/src/types.rs
+++ b/cardano-c/src/types.rs
@@ -100,8 +100,15 @@ impl From<txbuild::Error> for CardanoTransactionErrorCode {
 }
 
 #[repr(C)]
+pub enum DiffType {
+    Positive,
+    Negative,
+    Zero,
+}
+
+#[repr(C)]
 pub struct Balance {
-    sign: i32,
+    sign: DiffType,
     value: u64,
 }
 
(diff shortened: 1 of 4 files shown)


cardano-c/src/types.rs
```diff
@@ -82,6 +83,44 @@ impl CardanoTransactionErrorCode {
...
+            txbuild::Error::TxSignaturesMismatch => Self::signature_mismatch(),
+            txbuild::Error::CoinError(_) => Self::coin_out_of_bounds(),
+            txbuild::Error::FeeError(_) => unimplemented!(),
+        }
+    }
+}
+
+#[repr(C)]
+pub struct Balance {
+    sign: i32,
```
vincenthz: I would prefer to use an enum with a `repr(C)` and have the value on the C side as a C enum: `{ DIFF_POSITIVE = X, DIFF_NEGATIVE = X+1, DIFF_ZERO = X+2 }`

enzoc4: Something like this? https://github.com/input-output-hk/rust-cardano/pull/575/commits/dd65cf6cec7a6cb8581b18dfe448c85d4e5288b0

vincenthz: yes, thanks !

PR #589, opened 2019-03-28
[IOHKCRB-6] Example of how to create a wallet

enzoc4:
---
name: [IOHKCRB-6] Example of how to create a wallet
---

# Description

- Add examples/README.md with an example of how to create a wallet and an address, based on the one found in the [js-cardano-wasm](https://github.com/input-output-hk/js-cardano-wasm/tree/master/cardano-wallet ) repository.
- Update some documentation

### DIFF
--- a/cardano-c/examples/README.md
+++ b/cardano-c/examples/README.md
@@ -3,7 +3,6 @@
 This example shows how to create a wallet from english mnemonics
 
 ```C
-const char *alias = "Awesome Wallet";
 char *MNEMONICS = "crowd captain hungry tray powder motor coast oppose month shed parent mystery torch resemble index";
 
 /*Retrieve entropy from mnemonics*/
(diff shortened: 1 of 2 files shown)


cardano-c/examples/README.md
```diff
@@ -0,0 +1,45 @@
+# Creating a wallet
+
+This example shows how to create a wallet from english mnemonics
+
+```C
+const char *alias = "Awesome Wallet";
```
NicolasDP:
```suggestion
const char *alias = "Awesome Account";
```

It is the alias of an account, not a wallet. Just to make sure the readers are not confused but the wording

cardano-c/examples/README.md
```diff
@@ -0,0 +1,45 @@
...
+char *MNEMONICS = "crowd captain hungry tray powder motor coast oppose month shed parent mystery torch resemble index";
+
+/*Retrieve entropy from mnemonics*/
+cardano_entropy entropy;
+uint32_t bytes;
+cardano_bip39_error_t entropy_rc = cardano_entropy_from_english_mnemonics(MNEMONICS, &entropy, &bytes);
```
NicolasDP: can you also adds the way to check the returned error ? (I am happy with `assert(a == b);`)

cardano-c/examples/README.md
```diff
@@ -0,0 +1,50 @@
...
+/*Create an account*/
+const char *alias = "Awesome Account";
+unsigned int index = 0;
+cardano_account *account = cardano_account_create(wallet, alias, index);
+
+/*Create an internal address*/
+enum
+{
+    NUMBER_OF_ADDRESSES = 1,
+};
```
NicolasDP: why an enum here ?

enzoc4: It's one way to get a compilation time constant (because you can not use a variable for a statically stored array), but maybe a `#define` is better?. Would you prefer that?
Or something like using `1` in the declaration and just getting the size with `sizeof`?

Issue #581, opened 2019-03-27
mockchain: Block versioning tidy up


vincenthz:
at the moment block version (which define the format that we'll read) and the consensus algorithm is too intertwined. There's a need to keep a loose relation between versioning and consensus:

* `Block Version` defines `Consensus`
* `Consensus` defines which `Block Version` we accept

1) define an enum for the consensus that is unlinked to block version (current known values: BFT, GenesisPraos)
2) rename block version BFT to `block-ed25519-signed` (something related)
3) rename block version GenesisPraos to `block-kes-vrfproof` (something related)

PR #591, opened 2019-03-29
Block version clean-up

CodeSandwich:
Fixes https://github.com/input-output-hk/rust-cardano/issues/581

### DIFF
--- a/chain-impl-mockchain/src/block/version.rs
+++ b/chain-impl-mockchain/src/block/version.rs
@@ -1,32 +1,94 @@
+use std::collections::BTreeMap;
+use lazy_static::lazy_static;
 use num_derive::FromPrimitive;
 use num_traits::FromPrimitive;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct BlockVersion(pub(crate) u16);
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AnyBlockVersion {
+    Supported(BlockVersion),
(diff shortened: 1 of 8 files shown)


chain-impl-mockchain/src/leadership/genesis.rs
```diff
@@ -58,7 +58,7 @@ impl GenesisLeaderSelection {
 
     pub fn leader(
         &self,
-        vrf_key: &SecretKey<Curve25519_2HashDH>,
+        _vrf_key: &SecretKey<Curve25519_2HashDH>,
```
vincenthz: I agree that they those are annoying but shutting the warning when it's a WIP is not a good idea.

vincenthz: I agree that they those are annoying but shutting the warning when it's a WIP is not a good idea. In this case it also conflict with my next PR

chain-impl-mockchain/src/block/version.rs
```diff
@@ -1,32 +1,100 @@
...
+pub enum ConsensusVersion {
+    None = 0,
+    Bft = 1,
+    GenesisPraos = 2,
+}
 
-    pub fn from_block_version(ver: BlockVersion) -> Option<BlockVersionTag> {
-        BlockVersionTag::from_u16(ver.0)
+impl ConsensusVersion {
+    pub fn supported_block_versions(self) -> &'static [BlockVersion] {
```
vincenthz: whilst saving potential mismatch if you update one without the other, that seems a bit overkill in this case

PR #601, opened 2019-04-02
Multiverse improvements

edolstra:
This PR:

* Improves `Multiverse::gc()` by keeping all states with a chain length close to the current longest chain (as determined by the `SUFFIX_TO_KEEP` constant), as well as states at exponentially increasing gaps from the longest chain. This ensures that reconstructing an arbitrary state should be O(n) where n is the block's distance from the current longest chain.

* Adds a method `Multiverse::get_from_storage()` that gets the chain state at block `k` from memory if present, and otherwise reconstructs it by reading blocks from storage and applying them to the nearest ancestor state that we do have.

* Adds a test for `Multiverse`.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,4 +1,4 @@
-{ nixpkgs ? fetchTarball https://github.com/NixOS/nixpkgs/archive/7fff567ee99c1f343ecdd82fef2e35fb6f50e423.tar.gz
+{ nixpkgs ? fetchTarball https://github.com/NixOS/nixpkgs/archive/8d3e91077ba074e2c947a152ee8ab7be885c42ab.tar.gz
 , pkgs ? import nixpkgs {}
 }:
 
--- a/shell.nix
+++ b/shell.nix
@@ -5,7 +5,7 @@
 with pkgs;
 
 stdenv.mkDerivation {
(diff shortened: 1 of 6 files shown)


shell.nix
```diff
@@ -1,4 +1,4 @@
-{ nixpkgs ? fetchTarball https://github.com/NixOS/nixpkgs/archive/7fff567ee99c1f343ecdd82fef2e35fb6f50e423.tar.gz
+{ nixpkgs ? fetchTarball https://github.com/NixOS/nixpkgs/archive/8d3e91077ba074e2c947a152ee8ab7be885c42ab.tar.gz
 , pkgs ? import nixpkgs {}
 }:
```
NicolasDP: while you are changing this file, can you fix the line below:

```nix
stdenv.mkDerivation {	
  name = "rust-carano";	

  # ...
}
```

edolstra: Done!

vincenthz:
I didn't quite review everything, I think we need to discuss the design decision to make sure we are aware on how it works. overall it looks good and that it will fits, but I'm not sure about explicit GC roots instead of just relying of oldness (except maybe for old fork being populated while gc happens ?)

Issue #577, opened 2019-03-27
Check total value in the system at block0 time


vincenthz:
when reading block0 we need to make sure that the overall total value created for the system is less than the maximum value.

NicolasDP:
I guess the value would then go in the `LedgerStaticParameters`, right ?

PR #603, opened 2019-04-03
Add support for init message consensus

CodeSandwich:
This lets us save initial consensus with which genesis was created.

Also fixed https://github.com/input-output-hk/jormungandr/issues/239

Also fixes https://github.com/input-output-hk/rust-cardano/issues/577

### DIFF
--- a/chain-impl-mockchain/src/block/version.rs
+++ b/chain-impl-mockchain/src/block/version.rs
@@ -2,6 +2,8 @@ use lazy_static::lazy_static;
 use num_derive::FromPrimitive;
 use num_traits::FromPrimitive;
 use std::collections::BTreeMap;
+use strum::IntoEnumIterator;
+use strum_macros::{Display, EnumIter, EnumString, IntoStaticStr};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum AnyBlockVersion {
@@ -51,7 +53,7 @@ impl From<BlockVersion> for AnyBlockVersion {
     }
 }
 
-#[derive(Debug, Clone, Copy, FromPrimitive, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, EnumIter, FromPrimitive, PartialEq, Eq)]
 pub enum BlockVersion {
     Genesis = 0,
     Ed25519Signed = 1,
@@ -68,10 +70,26 @@ impl BlockVersion {
     }
 }
 
-#[derive(Debug, Clone, Copy, FromPrimitive, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(
+    Debug,
+    Clone,
+    Copy,
+    Display,
+    EnumString,
+    FromPrimitive,
+    IntoStaticStr,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+)]
 pub enum ConsensusVersion {
+    #[strum(to_string = "none")]
     None = 0,
+    #[strum(to_string = "bft")]
     Bft = 1,
+    #[strum(to_string = "genesis")]
     GenesisPraos = 2,
 }
 
@@ -80,14 +98,10 @@ impl ConsensusVersion {
         lazy_static! {
             static ref MAPPING: BTreeMap<u16, Vec<BlockVersion>> = {
                 let mut map = BTreeMap::<_, Vec<_>>::new();
-                for block_ord in 0.. {
-                    match BlockVersion::from_u64(block_ord) {
-                        Some(block) => map
-                            .entry(block.get_consensus() as u16)
-                            .or_default()
-                            .push(block),
-                        None => break,
-                    }
+                for block_version in BlockVersion::iter() {
+                    map.entry(block_version.get_consensus() as u16)
+                        .or_default()
+                        .push(block_version)
                 }
                 map
             };
--- a/chain-impl-mockchain/src/block/version.rs
+++ b/chain-impl-mockchain/src/block/version.rs
@@ -61,11 +61,11 @@ pub enum BlockVersion {
 }
 
 impl BlockVersion {
-    pub fn get_consensus(self) -> ConsensusVersion {
+    pub fn get_consensus(self) -> Option<ConsensusVersion> {
         match self {
-            BlockVersion::Genesis => ConsensusVersion::None,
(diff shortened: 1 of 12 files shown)


chain-impl-mockchain/src/block/version.rs
```diff
@@ -68,10 +70,25 @@ impl BlockVersion {
...
+    PartialOrd,
+    Ord,
+    Hash,
+)]
+#[strum(serialize_all = "snake_case")]
 pub enum ConsensusVersion {
     None = 0,
     Bft = 1,
+    #[strum(serialize = "genesis")]
     GenesisPraos = 2,
```
NicolasDP: I am not sure `Strum` was needed just for this. Worse, having an automatically generated value for what is in the block does not inspire me confidence. I'd rather like to have the code written hand here (especially it is rather simple here)

CodeSandwich: We need both parsing and printing textual representation. `Strum` lets the implementation be DRY, it's impossible to make mistake now and we have less boilerplate. It's exactly like `num_derive`, which we already use, but for strings.

You're right about generated values being dangerously cryptic. How about explicitly defining every variant representation like in `GenesisPraos`?

CodeSandwich:
PLEASE DO NOT MERGE
I've found a place where I haven't added support for consensus version. I think I should make ConfigParam a real enum, so compiler can warn us about such situations.

chain-impl-mockchain/src/block/version.rs
```diff
@@ -68,10 +70,26 @@ impl BlockVersion {
...
+    FromPrimitive,
+    IntoStaticStr,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+)]
 pub enum ConsensusVersion {
+    #[strum(to_string = "none")]
```
NicolasDP: is there a way to make this value not possible in the string? None is only valid when the system starts up, but it is not a valid value in the initial message

CodeSandwich: Is it valid in genesis.yaml?

NicolasDP: No it is not.

CodeSandwich: I'm starting to suspect that the design of ConsensusVersion may be broken. Is `None` variant valid in any context? If system starts up and the inintial message does not set consensus version of block0, the node is in state not accepted by multiple mechanisms.

Maybe the None variant should be removed completely and block0-consensus in initial message should be mandatory (either bft of genesis)?

NicolasDP: Yes, it seems you are right. None should be removed here. The value is mandatory.

chain-impl-mockchain/src/transaction/mod.rs
```diff
@@ -25,8 +25,13 @@ impl<Extra: property::Serialize> property::Serialize for AuthenticatedTransactio
     type Error = Extra::Error;
 
     fn serialize<W: std::io::Write>(&self, mut writer: W) -> Result<(), Extra::Error> {
-        assert_eq!(self.transaction.inputs.len(), self.witnesses.len());
-
+        if self.transaction.inputs.len() != self.witnesses.len() {
```
vincenthz: This shouldn't be enforced by the serialization layer, please remove completely

chain-impl-mockchain/src/ledger.rs
```diff
@@ -156,7 +156,7 @@ impl Ledger {
                 params.block0_initial_hash = block0_hash;
                 Ok(params)
             }
-            Some(_) => Err(Error::Block0InitialMessageMissing),
+            _ => Err(Error::Block0InitialMessageMissing),
```
vincenthz: no, the error message is suppose to be different in the first place (I used the already existing error when adding this big, and forgot to add the new error). in the case of `Some(_)` it should be a `InitialMessageExpectedButGotSomethingElse`

chain-impl-mockchain/src/ledger.rs
```diff
@@ -299,6 +300,14 @@ impl Ledger {
     pub fn chain_length(&self) -> ChainLength {
         self.chain_length
     }
+
+    fn validate_utxo_total_value(&self) -> Result<(), Error> {
+        let old_utxo_values = self.oldutxos.iter().map(|entry| entry.output.value);
+        let new_utxo_values = self.utxos.iter().map(|entry| entry.output.value);
```
vincenthz: you also need to sum the coins from self.accounts

chain-impl-mockchain/src/ledger.rs
```diff
@@ -148,11 +150,13 @@ impl Ledger {
...
                         }
                         ConsensusVersion::TAG => {
-                            params.block0_consensus = ConsensusVersion::from_payload(payload)?
+                            consensus = Some(ConsensusVersion::from_payload(payload)?)
                         }
                         _ => return Err(Error::UnknownConfig(*tag)),
                     }
                 }
+                params.block0_consensus =
+                    consensus.ok_or(Error::Block0InitialMessageNoConsensus)?;
```
vincenthz: nice !

chain-impl-mockchain/src/config.rs
```diff
@@ -99,75 +205,102 @@ impl ConfigParam for Discrimination {
...
+    fn from_cfg_str(s: &str) -> Result<Self, Error> {
         s.parse().map_err(|_| Error::UnknownString(s.to_string()))
     }
 }
 
-pub fn entity_to<T: ConfigParam>(t: &T) -> (Tag, TagPayload) {
-    (T::TAG, t.to_payload())
-}
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+struct Tag(u16);
```
vincenthz: it would be better not to have moved code around (Tag/TagLen), it make reading the diff a bit harder.

chain-impl-mockchain/src/transaction/mod.rs
CodeSandwich: I've moved this check to JCLI

chain-impl-mockchain/src/config.rs
CodeSandwich: Tag and TagLen became private, so I wanted them not to overshadow public things, which are obviously more important for readers. I considered moving them to a private module in separate file to make project easier to navigate, but the team seems to prefer long, self-contained files.

PR #612, opened 2019-04-06
Communicate node ID in metadata of subscription requests

mzabaluev:
As the gRPC protocol is stateless, and transport connection information cannot reliably be used to establish the peer identity, we need to exchange node IDs along with subscription requests. Request metadata is the place to do it.

The sender ID field from `Gossip` therefore becomes redundant and is removed.

### DIFF
--- a/network-core/src/gossip.rs
+++ b/network-core/src/gossip.rs
@@ -23,23 +23,17 @@ pub trait Node: Serialize + Deserialize {
 
 #[derive(Clone, Debug)]
 pub struct Gossip<T: Node> {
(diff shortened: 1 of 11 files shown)


mzabaluev:
Ugh, I meant to publish this as a draft. The IDs are not yet sent on the client side. Stay tuned for more commits.

NicolasDP:
If you want, there is a _label_ `X-WIP` you can use to warn others it is not meant to merge already

mzabaluev:
Rebased and ready now.

PR #613, opened 2019-04-06
add time referencial

vincenthz:


chain-time/src/timeframe.rs
```diff
@@ -0,0 +1,168 @@
...
+                Some(SlotAndDuration {
+                    slot: Slot(self.slot_offset.0 + slot_nb),
+                    slot_duration_offset: d,
+                })
+            }
+        }
+    }
+
+    /// Given a system time get the slot associated
+    pub fn to_slot(&self, at: &SystemTime) -> Option<Slot> {
```
mzabaluev: The naming of this and the above method is a bit misleading. `to_` is normally used for one-to-other conversions, while this uses system time and self as the basis.
I think `slot_at` would be a better name.

chain-time/src/timeframe.rs
```diff
@@ -0,0 +1,168 @@
...
+    ///                   |
+    ///                   frame.change_frame(Slot(2), NbSeconds(4))
+    /// ```
+    ///
+    pub fn change_frame(&self, slot: Slot, duration_per_slot: NbSeconds) -> Self {
+        let d = Duration::from_secs(slot.0 * self.slot_duration.0);
+        let new_timeline = self.timeline.advance(d);
+        TimeFrame {
+            timeline: new_timeline,
+            slot_offset: Slot(self.slot_offset.0 + slot.0),
```
mzabaluev: Consider making a `Slot::offset` method which would use checked math under the hood.

chain-time/src/timeline.rs
```diff
@@ -0,0 +1,37 @@
...
+        match t.duration_since(self.0) {
+            Ok(duration) => Some(TimeOffset(duration)),
+            Err(_) => None,
+        }
+    }
+
+    /// Advance a timeline, and create a new timeline starting at
+    /// timeline + duration
+    pub fn advance(&self, d: Duration) -> Self {
+        Timeline(self.0 + d)
```
mzabaluev: Unfortunately `time_checked_add` is not a stable Rust feature, so this should go for now.

chain-time/src/timeframe.rs
```diff
@@ -0,0 +1,168 @@
...
+impl From<Slot> for u64 {
+    fn from(s: Slot) -> u64 {
+        s.0
+    }
+}
+
+/// Identify a slot in a specific timeframe and a leftover duration
+#[derive(Debug)]
+pub struct SlotAndDuration {
+    pub slot: Slot,
```
mzabaluev: Consider hiding the data members and providing read-only accessors instead.

chain-time/src/timeframe.rs
```diff
@@ -0,0 +1,168 @@
...
+/// Time frame which is a timeline that is configured to be split in discrete slots
+pub struct TimeFrame {
+    timeline: Timeline,
+    slot_offset: Slot,
+    slot_duration: NbSeconds,
+    //slot_configuration: Vec<u64>,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct NbSeconds(u64);
```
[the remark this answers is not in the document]
mzabaluev: Maybe just rename it `SlotDuration`?

chain-time/src/timeframe.rs
```diff
@@ -0,0 +1,168 @@
...
+
+impl NbSeconds {
+    pub fn new(seconds: u32) -> Self {
```
mzabaluev: Continuing with the above, naming the constructor `SlotDuration::from_secs` would clarify the input units and match with `Duration::from_secs`. The internal representation is not public.

chain-time/src/timeframe.rs
```diff
@@ -0,0 +1,168 @@
...
+    ///                   |
+    ///                   ┕---x---x---x---x---x         returned frame
+    ///                   2   3   4   5   6   7
+    ///                   ↑
+    ///                   |
+    ///                   frame.change_frame(Slot(2), NbSeconds(4))
+    /// ```
+    ///
+    pub fn change_frame(&self, slot: Slot, duration_per_slot: NbSeconds) -> Self {
+        let d = Duration::from_secs(slot.0 * self.slot_duration.0);
```
mzabaluev: Consider using checked arithmetics.

chain-time/src/timeframe.rs
```diff
@@ -0,0 +1,168 @@
...
+
+impl From<Slot> for u64 {
+    fn from(s: Slot) -> u64 {
+        s.0
+    }
+}
+
+/// Identify a slot in a specific timeframe and a leftover duration
+#[derive(Debug)]
+pub struct SlotAndDuration {
```
mzabaluev: I would bikeshed this as a more concise `SlotTime`.

chain-time/src/timeframe.rs
```diff
@@ -0,0 +1,168 @@
...
+    pub slot: Slot,
+    /// The offset of a specific time frame in
+    pub slot_duration_offset: Duration,
```
mzabaluev: Just `offset` (as a reader method) would suffice.

mzabaluev: The type of this might want to be `SlotDuration` for extra checking, but maybe convenience suggests otherwise.

chain-time/src/timeline.rs
```diff
@@ -0,0 +1,37 @@
...
+    /// Create a new timeline, which is a time starting point
+    pub fn new(start_time: SystemTime) -> Self {
+        Timeline(start_time)
+    }
+
+    /// Return the duration since the creation of the timeline
+    ///
+    /// If the time is earlier than the start of this timeline,
+    /// then None is returned.
+    pub fn differential(&self, t: &SystemTime) -> Option<TimeOffset> {
```
mzabaluev: I wanted to bikeshed this method name but then I found out it is [Doctor Who compliant](https://tardis.fandom.com/wiki/Time_differential). Approved.

chain-time/src/timeframe.rs
```diff
@@ -0,0 +1,168 @@
...
+        let new_timeline = self.timeline.advance(d);
+        TimeFrame {
+            timeline: new_timeline,
+            slot_offset: Slot(self.slot_offset.0 + slot.0),
+            slot_duration: duration_per_slot,
+        }
+    }
+
+    /// Given a system time get the slot and associated duration leftover
+    pub fn to_slot_and_remaining(&self, at: &SystemTime) -> Option<SlotAndDuration> {
```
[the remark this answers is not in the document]
mzabaluev: In theme with earlier naming suggestions, this would be `slot_time_at`.

chain-time/src/timeframe.rs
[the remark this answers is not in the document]
vincenthz: good idea

chain-time/src/timeframe.rs
vincenthz: Done

chain-time/src/timeframe.rs
vincenthz: this and the other calculation, I don't really see a point of using checked math at this moment. it does comes with a high prices in term of API for a corner case that we shouldn't really encounter

chain-time/src/timeframe.rs
vincenthz: I'm not a fan of either name, so leaving it be for now, until it get some service anyway

chain-time/src/timeframe.rs
vincenthz: I've changed the name, but no I spefically want the extra precision from Duration (think nanosecs) here

chain-time/src/timeframe.rs
mzabaluev: Does this provide much added value over a `(Slot, Duration)` tuple?

PR #614, opened 2019-04-06
add serde generic serialization as an optional feature

NicolasDP:
also update the CI to test the new added serde implementations

here, serde is added as an optional dependency. It not meant to be used to store the blockchain in a generic format (this would be a bad design choice as it will certainly store excessive metadata and will end up taking more and more storage).

The serde instances are used to standardise some interfaces that we will have on the side:

* REST API (json, ...);
* configuration files (yaml, ...);
* inter process (bincode, ...).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,49 +8,156 @@ jobs:
       - checkout
       - restore_cache:
           keys:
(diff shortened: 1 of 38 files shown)


NicolasDP:
closing this PR, this is a mistake to add serde to the low level library.

PR #616, opened 2019-04-08
update CircleCI

NicolasDP:


.circleci/config.yml
```diff
@@ -8,146 +8,166 @@ jobs:
...
+      - run: git submodule update --init
       - run: cargo fetch
       - persist_to_workspace:
-          root: "."
+          root: /mnt/crate
           paths:
             - Cargo.lock
       - save_cache:
-          key: cargo-v2-{{ checksum "Cargo.toml" }}-{{ checksum "Cargo.lock" }}
+          key: cargo-v3-{{ checksum "Cargo.toml" }}
```
mzabaluev: The point of doing it the old way was that the cache does not need to be re-saved if `Cargo.lock` did not change. `restore_cache` makes sure partial matches are considered.

.circleci/config.yml
```diff
@@ -8,146 +8,166 @@ jobs:
...
       - checkout
       - attach_workspace:
-          at: "."
+          at: /mnt/crate
+      - run: git submodule sync
+      - run: git submodule update --init
       - restore_cache:
           keys:
-            - cargo-v2-{{ checksum "Cargo.toml" }}-{{ checksum "Cargo.lock" }}
+            - cargo-v3-{{ checksum "Cargo.toml" }}-{{ checksum "Cargo.lock" }}
```
mzabaluev: Now this line never matches because the key never includes the `Cargo.lock` checksum?

PR #623, opened 2019-04-09
chain-addr: allow configurable address prefixes

sjmackenzie:
set by editing the relevant environment variables:

`$ TESTING_ADDRESS_PREFIX=teo PRODUCTION_ADDRESS_PREFIX=ceo cargo test`

The address tests will inevitably fail as they are configured
for ca & ta prefixes. This is an acceptable tradeoff.

### DIFF
--- a/chain-addr/build.rs
+++ b/chain-addr/build.rs
@@ -7,6 +7,9 @@ fn main() {
     let production_prefix = option_env!("PRODUCTION_ADDRESS_PREFIX").unwrap_or("ca");
     let test_prefix = option_env!("TEST_ADDRESS_PREFIX").unwrap_or("ta");
 
-    println!("cargo:rustc-env=PRODUCTION_ADDRESS_PREFIX={}", production_prefix);
+    println!(
+        "cargo:rustc-env=PRODUCTION_ADDRESS_PREFIX={}",
+        production_prefix
+    );
     println!("cargo:rustc-env=TEST_ADDRESS_PREFIX={}", test_prefix);
 }
--- a/chain-addr/build.rs
+++ b/chain-addr/build.rs
@@ -1,0 +1,12 @@
+use std::env;
+use std::fs::File;
+use std::io::Write;
+use std::path::Path;
+
+fn main() {
+    let production_prefix = option_env!("PRODUCTION_ADDRESS_PREFIX").unwrap_or("ca");
+    let test_prefix = option_env!("TEST_ADDRESS_PREFIX").unwrap_or("ta");
+
+    println!("cargo:rustc-env=PRODUCTION_ADDRESS_PREFIX={}", production_prefix);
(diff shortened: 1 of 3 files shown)


chain-addr/build.rs
```diff
@@ -0,0 +1,16 @@
...
+    let out_dir = env::var("OUT_DIR").expect("No out dir");
+    let dest_path = Path::new(&out_dir).join("address_constants.rs");
+    let mut f = File::create(&dest_path).expect("Could not create file");
+
+    let production_prefix = option_env!("PRODUCTION_ADDRESS_PREFIX").unwrap_or("ca");
+    let test_prefix = option_env!("TEST_ADDRESS_PREFIX").unwrap_or("ta");
+
+    write!(&mut f, "const PRODUCTION_ADDRESS_PREFIX: &'static str = \"{}\";\n", production_prefix).expect("Could not write file");
+    write!(&mut f, "const TEST_ADDRESS_PREFIX: &'static str = \"{}\";\n", test_prefix).expect("Could not write file");
+}
```
NicolasDP: I don't like the idea of adding a file here. Utilising the cargo env would be much simpler.

have a look at this: https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script

I think instead:

```rust
    let production_prefix = option_env!("PRODUCTION_ADDRESS_PREFIX").unwrap_or("ca");
    let test_prefix = option_env!("TEST_ADDRESS_PREFIX").unwrap_or("ta");

    println!("cargo:rustc-env=PRODUCTION_ADDRESS_PREFIX={}", production_prefix);
    println!("cargo:rustc-env= TEST_ADDRESS_PREFIX ={}", test_prefix);
```

And in the chain-addr/src/lib.rs

```rust
const PRODUCTION_ADDRESS_PREFIX: &'static str = env!("PRODUCTION_ADDRESS_PREFIX");
const TEST_ADDRESS_PREFIX: &'static str = env!("TEST_ADDRESS_PREFIX");
```

chain-addr/build.rs
```diff
@@ -0,0 +1,12 @@
+use std::env;
+use std::fs::File;
+use std::io::Write;
+use std::path::Path;
+
+fn main() {
+    let production_prefix = option_env!("PRODUCTION_ADDRESS_PREFIX").unwrap_or("ca");
+    let test_prefix = option_env!("TEST_ADDRESS_PREFIX").unwrap_or("ta");
+
+    println!("cargo:rustc-env=PRODUCTION_ADDRESS_PREFIX={}", production_prefix);
```
NicolasDP: please, apply rustfmt or do the following change:

```diff
-    println!("cargo:rustc-env=PRODUCTION_ADDRESS_PREFIX={}", production_prefix);
+    println!(
+        "cargo:rustc-env=PRODUCTION_ADDRESS_PREFIX={}",
+        production_prefix
+    );
```

Issue #610, opened 2019-04-05
Complete initial state


vincenthz:
Here's a somewhat longer list of additional config parameters:

* Slot number per epoch
* Slot durations
* Stability
* Consensus requires a bunch of new extra parameters:
  * BFT:
    * a set of Bft `LeaderId`
  * Genesis:
    * parameter D (between 0.0 and 1.0)
    * parameter F (between 0.0 and 1.0)
    * note the Bft `LeaderId` is also useful in Genesis

I think there's a bunch of other parameters that are not on this list, but the big picture here is to have the blockchain fully configured through this message.

As additional side stuff to look into it:
* how to encode "float". likely we don't want ieee754 encoding, but a *decimal* based encoding with a specific precision and a specific range (0.0 to 1.0)
* macro(s) (if possible) to quickly add new parameter and their general encoding

CodeSandwich:
@vincenthz The addition of consensus parameters from `genesis.yaml` perspective:
```yaml
# Current design
# problem: no way to have parameters in consensus 
blockchain_configuration:
  - [ block0-date, "1687616871" ]
  - [ discrimination, test ]
  - [ block0-consensus, genesis ]

# Plan A
# the forceful extension of current design, needs custom parsing,
# breaks idea of YAML
blockchain_configuration:
  - [ block0-date, "1687616871" ]
  - [ discrimination, test ]
  - [ block0-consensus, genesis:500:500:ad75ad875f9708730457cb98 ]

# Plan B
# the dreaded YAML format change, needs refactoring so config items can
# deserialize themselves without relying on blockchain_configuration knowing how to
# deserialize each of them (which I think is easy and healthy for the code anyway)
blockchain_configuration:
  - block0-date: "1687616871"
  - discrimination: test
  - block0-consensus:
    kind: genesis,
    param_d: 500,
    param_f: 500,
    leader: ad75ad875f9708730457cb98

# Plan C
# the multiplication of items, makes either deserialization or usage tricky,
# because data is not groupped, items have to be composed from
# multiple entries and verified for consistency
blockchain_configuration:
  - [ block0-date, "1687616871" ]
  - [ discrimination, test ]
  - [ block0-consensus, genesis:500:500:ad75ad875f9708730457cb98 ]
  - [ block0-consensus_param_d, 500 ]
  - [ block0-consensus_param_f, 500 ]
  - [ block0-consensus_leader, ad75ad875f9708730457cb98 ]
```
[...]

PR #633, opened 2019-04-11
Extend init message config params

CodeSandwich:
This adds functionality required by https://github.com/input-output-hk/rust-cardano/issues/610

### DIFF
--- a/chain-impl-mockchain/src/config.rs
+++ b/chain-impl-mockchain/src/config.rs
@@ -1,16 +1,16 @@
 use crate::block::ConsensusVersion;
+use crate::leadership::bft::LeaderId;
+use crate::milli::Milli;
 use chain_addr::Discrimination;
 use chain_core::mempack::{ReadBuf, ReadError, Readable};
 use chain_core::packer::Codec;
 use chain_core::property;
+use chain_crypto::PublicKey;
+use num_derive::FromPrimitive;
 use num_traits::FromPrimitive;
 use std::fmt::{self, Display, Formatter};
 use std::io::{self, Write};
-use std::str::FromStr;
-
-/// Seconds elapsed since 1-Jan-1970 (unix time)
-#[derive(Debug, Clone, Copy, Eq, PartialEq)]
-pub struct Block0Date(pub u64);
+use strum_macros::{AsRefStr, EnumIter, EnumString};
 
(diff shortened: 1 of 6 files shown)


chain-impl-mockchain/src/leadership/bft.rs
```diff
@@ -105,6 +106,18 @@ impl From<PublicKey<SIGNING_ALGORITHM>> for LeaderId {
     }
 }
 
+impl Bech32 for LeaderId {
```
NicolasDP: I am not sure this is wise to start implementing this trait for LeaderId. There is no reason to believe this type and a public key are both the same object. The HRP here is not necessarily true.

chain-impl-mockchain/src/config.rs
```diff
@@ -224,24 +287,102 @@ impl ConfigParamVariant for ConsensusVersion {
...
-        // validate that it's less than 1024 when const fns support ifs
-        Tag(tag)
+    fn from_payload(payload: &[u8]) -> Result<Self, Error> {
+        PublicKey::from_binary(payload)
+            .map(Into::into)
+            .map_err(|_| Error::SizeInvalid)
+    }
+
+    fn to_cfg_string(&self) -> String {
+        self.as_public_key().to_bech32_str()
```
NicolasDP: I am not sure the encoding in `bech32` holds. Actually this comes to: why this library defines the encoding format as it should be in the `jcli`'s genesis.yaml file, which is application defined.

It seems to me the `to_cfg_string` and `from_cfg_str` functions are not needed at all here.

chain-impl-mockchain/src/config.rs
```diff
@@ -93,84 +153,89 @@ mod serde_impl {
 
     impl<'de> Deserialize<'de> for ConfigParam {
```
NicolasDP: Please, remove this module `serde_impl`. This is never compiled and will never be.

chain-impl-mockchain/src/config.rs
CodeSandwich: I was thinking about removing `to_cfg_string` and `from_cfg_str` completely. It's only used by `ConfigParam`'s Serde implementation, which we want to get rid of anyway. I guess it's a good moment to purge it.

PR #635, opened 2019-04-15
Remove boxing and dynamic dispatch from traits featuring iterators

mzabaluev:
Found another way to specify `HasMessages` that does not require allocations or dynamic dispatch.
An internal trait in mockchain that used boxed trait object has also been simplified to just return slice references.


vincenthz:
that does looks better, should we merge this or is there's something preventing merging ? (apart from the now recently introduced merge conflict)

mzabaluev:
Rebased; trait `HasPublicKeys` required more work this time.

PR #637, opened 2019-04-16
Remove non-canonical usage of Deref

CodeSandwich:
Fixes https://github.com/input-output-hk/rust-cardano/issues/329

### DIFF
--- a/cardano/src/block/normal.rs
+++ b/cardano/src/block/normal.rs
@@ -36,7 +36,7 @@ impl BodyProof {
 
     pub fn generate_from_body(body: &Body) -> Self {
         BodyProof {
-            tx: tx::TxProof::generate(&body.tx),
+            tx: tx::TxProof::generate(&body.tx.txaux),
             mpc: SscProof::generate(&body.ssc),
             delegation: DlgProof::generate(&body.delegation),
             update: update::UpdateProof::generate(&body.update),
@@ -104,8 +104,9 @@ impl cbor_event::de::Deserialize for DlgProof {
 
 #[derive(Debug, Clone)]
 pub struct TxPayload {
-    txaux: Vec<tx::TxAux>,
+    pub txaux: Vec<tx::TxAux>,
 }
+
 impl fmt::Display for TxPayload {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         if self.txaux.is_empty() {
@@ -117,6 +118,7 @@ impl fmt::Display for TxPayload {
         write!(f, "")
     }
 }
+
 impl TxPayload {
     pub fn new(txaux: Vec<tx::TxAux>) -> Self {
         TxPayload { txaux: txaux }
@@ -128,6 +130,7 @@ impl TxPayload {
         self.txaux.iter()
     }
 }
+
 impl IntoIterator for TxPayload {
     type Item = <Vec<tx::TxAux> as IntoIterator>::Item;
     type IntoIter = <Vec<tx::TxAux> as IntoIterator>::IntoIter;
@@ -135,17 +138,7 @@ impl IntoIterator for TxPayload {
         self.txaux.into_iter()
     }
 }
-impl ::std::ops::Deref for TxPayload {
-    type Target = Vec<tx::TxAux>;
-    fn deref(&self) -> &Self::Target {
-        &self.txaux
-    }
-}
-impl ::std::ops::DerefMut for TxPayload {
-    fn deref_mut(&mut self) -> &mut Vec<tx::TxAux> {
-        &mut self.txaux
-    }
-}
+
 impl cbor_event::se::Serialize for TxPayload {
     fn serialize<'se, W: Write>(
         &self,
@@ -154,6 +147,7 @@ impl cbor_event::se::Serialize for TxPayload {
         cbor_event::se::serialize_indefinite_array(self.txaux.iter(), serializer)
     }
 }
+
 impl cbor_event::de::Deserialize for TxPayload {
     fn deserialize<R: BufRead>(raw: &mut Deserializer<R>) -> cbor_event::Result<Self> {
         let num_inputs = raw.array()?;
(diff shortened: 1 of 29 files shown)


cardano/src/block/normal.rs
```diff
@@ -104,8 +104,9 @@ impl cbor_event::de::Deserialize for DlgProof {
 
 #[derive(Debug, Clone)]
 pub struct TxPayload {
-    txaux: Vec<tx::TxAux>,
+    pub txaux: Vec<tx::TxAux>,
```
mzabaluev: I'd rather not expose internal members. In fact, I don't understand the purpose of this type which is just a wrapper over `Vec<tx::TxAux>`. Perhaps more principled accessor methods are needed on `Tx`, such as:

```rust
    fn contents(&self) -> &[tx::TxAux] {
        // ...
    }
```

cardano/src/block/normal.rs
```diff
@@ -543,7 +537,8 @@ impl cbor_event::de::Deserialize for DecShare {
 // BTreeMap<StakeholderId, VssCertificate> see
 // https://github.com/input-output-hk/cardano-sl/blob/005076eb3434444a505c0fb150ea98e56e8bb3d9/core/src/Pos/Core/Ssc/VssCertificatesMap.hs#L36-L44
 #[derive(Debug, Clone)]
-pub struct VssCertificates(Vec<VssCertificate>);
+pub struct VssCertificates(pub Vec<VssCertificate>);
```
mzabaluev: Same thing here; why is it even a newtype?

cardano/src/block/verify.rs
```diff
@@ -323,18 +323,18 @@ impl Verify for VssCertificates {
     fn verify(&self, protocol_magic: ProtocolMagic) -> Result<(), Error> {
         // check that there are no duplicate VSS keys
         let mut vss_keys = BTreeSet::new();
-        if !self.iter().all(|x| vss_keys.insert(x.vss_key.clone())) {
+        if !self.0.iter().all(|x| vss_keys.insert(x.vss_key.clone())) {
```
mzabaluev: I don't like all these `.0` accessors. If a wrapper type has a purpose (i.e. it's not just a meaningless newtype over something that everybody would actually want to use anyway), it should hide its contents and provide a named accessor method or methods to the inner value, `as`-style if accessing by reference, `into`-style if consumed to get the inner object.

cardano/src/block/verify.rs
```diff
@@ -616,7 +616,7 @@ mod tests {
             if let Block::MainBlock(mblk) = &mut blk {
                 match &mut mblk.body.ssc {
                     normal::SscPayload::CommitmentsPayload(_, vss_certs) => {
-                        vss_certs[0].expiry_epoch = 123;
+                        vss_certs.0[0].expiry_epoch = 123;
```
mzabaluev: Ugh.

cardano/src/util/arbitrary.rs
```diff
@@ -75,20 +71,20 @@ impl Arbitrary for Wrapper<(hdwallet::XPrv, address::ExtendedAddr)> {
...
             derivation_path: None,
             stake_distribution: address::StakeDistribution::BootstrapEraDistr,
-            network_magic: *network_magic,
+            network_magic: network_magic.into_inner(),
         };
         let addr_type = address::AddrType::ATPubKey;
         let addr = address::HashedSpendingData::new(
             addr_type,
-            &address::SpendingData::PubKeyASD(xprv.public()),
+            &address::SpendingData::PubKeyASD(xprv.clone().into_inner().public()),
```
mzabaluev: It's more performant to have a by-ref accessor here and clone the inner object: `xprv.wotsit().clone()`

cardano/src/wallet/bip44.rs
```diff
@@ -486,38 +463,36 @@ impl ChangeLevel<XPub> {
...
 }
-impl<T> Deref for ChangeLevel<T> {
-    type Target = T;
-    fn deref(&self) -> &T {
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct IndexLevel<T>(T);
+
+impl<T> IndexLevel<T> {
+    pub fn get_key(&self) -> &T {
```
mzabaluev: Please be Rust style compliant and name the getter method just `key`.

cardano/src/wallet/rindex.rs
```diff
@@ -382,16 +377,14 @@ impl RootKey {
         self.root_key
     }
 
+    pub fn get_xprv(&self) -> &XPrv {
```
mzabaluev: Same here, `xprv`.

chain-impl-mockchain/src/multiverse.rs
```diff
@@ -62,12 +62,9 @@ impl GCRoot {
 
         GCRoot { hash, roots }
     }
-}
 
-impl std::ops::Deref for GCRoot {
-    type Target = BlockId;
-    fn deref(&self) -> &Self::Target {
-        &self.hash
+    pub fn get_hash(&self) -> BlockId {
```
mzabaluev: Just `hash`, or maybe `id`?

exe-common/src/config.rs
```diff
@@ -234,17 +223,17 @@ pub mod net {
...
+                .iter()
                 .filter_map(|np| np.peer().get_native())
                 .collect()
         }
-    }
-    impl Deref for Peers {
-        type Target = Vec<NamedPeer>;
-        fn deref(&self) -> &Self::Target {
+
+        pub fn get_peers(&self) -> &[NamedPeer] {
```
mzabaluev: `peers`

exe-common/src/network/native.rs
```diff
@@ -96,34 +93,30 @@ impl Api for PeerPool {
...
-impl Deref for Connection {
-    type Target = OpenPeer;
-    fn deref(&self) -> &Self::Target {
-        &self.1
-    }
-}
-impl DerefMut for Connection {
-    fn deref_mut(&mut self) -> &mut Self::Target {
+
+    pub fn get_peer_mut(&mut self) -> &mut OpenPeer {
```
mzabaluev: `peer_mut`

vincenthz:
closing since it doesn't apply anymore, and it would make sense to split this into smaller / more digestable piece. At the moment, I think we should all keep this in mind in the new code and prevent `Deref` to be added needlessly. Changing the actual interface in `cardano` is somewhat painful (multi repository to check/adjust), so it might be better to do this when we have time for janitorial work

PR #638, opened 2019-04-16
add Display implementation for the stake pool id

NicolasDP:


chain-impl-mockchain/src/stake/role.rs
```diff
@@ -141,6 +141,11 @@ impl From<chain_crypto::Blake2b256> for StakePoolId {
         StakePoolId(hash.into())
     }
 }
+impl std::fmt::Display for StakePoolId {
```
mzabaluev: Could reduce path noise with importing `std::fmt` at the top, looks good otherwise.

PR #639, opened 2019-04-16
[IOHKCRB-21] Create C functions to process cbor (raw) blocks

enzoc4:
---
name: [IOHKCRB-21] Create C functions to process cbor blocks
about: ^

---

# Description
The new functions work on opaque pointers over the already existing abstractions, to keep consistency and code reusability. For example, the function `cardano_block_get_transactions` returns an array of pointers to `cardano_signed_transaction`, because of this, it was necessary to add accesors, for example `cardano_transaction_signed_get_outputs`.

To avoid copying data all the time, the functions that return pointers work as "references", for example: `cardano_block_get_transactions` has pointers to the data inside the block, and so, this data is only valid while `cardano_block_delete` is not called on the corresponding block. 
### Decoder function
- `cardano_raw_block_decode`
- `cardano_block_delete`
### Block accesor
- `cardano_block_get_transactions`
- `cardano_block_delete_transactions`
### TxAux accesor
- `cardano_transaction_signed_get_outputs`
- `cardano_transaction_signed_delete_outputs`
- `cardano_transaction_signed_get_inputs`
- `cardano_transaction_signed_delete_inputs`

### TxOutput accesors
- `cardano_transaction_txoutput_address`
- `cardano_transaction_txoutput_value`

### TxoPtr accesors
- `cardano_transaction_txoptr_index`
- `cardano_transaction_txoptr_txid`

## Example
```C
cardano_block *block;
cardano_raw_block_decode(RAW_BLOCK, RAW_BLOCK_SIZE, &block);
size_t transactions_size;
cardano_signed_transaction **transactions;
cardano_result rc = cardano_block_get_transactions(block, &transactions, &transactions_size);
assert(rc != CARDANO_RESULT_ERROR);
assert(size > 0);

cardano_txoptr **inputs;
size_t inputs_size;

cardano_transaction_signed_get_inputs(transactions[0], &inputs, &inputs_size);

cardano_txid_t txid;
cardano_transaction_txoptr_txid(inputs[0], &txid);

cardano_transaction_signed_delete_inputs(inputs, inputs_size);
cardano_block_delete_transactions(transactions, transactions_size);
cardano_block_delete(block);

```


cardano-c/cardano.h
```diff
@@ -251,8 +252,53 @@ typedef struct cardano_txoptr cardano_txoptr;
...
+*/
+void cardano_transaction_signed_get_outputs(
+    cardano_signed_transaction *txaux,
+    cardano_txoutput *(*outputs[]),
+    size_t *outputs_size
+);
+
+/*! \brief Release the memory allocated by `cardano_transaction_signed_get_outputs`
+*/
+void cardano_transaction_signed_delete_outputs(cardano_txoutput *outputs[], size_t size);
```
NicolasDP:
```suggestion
void cardano_signed_transaction_delete_outputs(cardano_txoutput *outputs[], size_t size);
```

cardano-c/src/block.rs
```diff
@@ -0,0 +1,69 @@
...
+    size: usize,
+    out: *mut BlockPtr,
+) -> CardanoResult {
+    let slice = unsafe { std::slice::from_raw_parts(bytes, size) };
+    let raw_block = block::block::RawBlock::from_dat(slice.to_vec());
+    let block = match raw_block.decode() {
+        Ok(b) => b,
+        Err(_) => return CardanoResult::failure(),
+    };
+    let pointer = Box::into_raw(Box::new(block));
```
NicolasDP: Why not?

```suggestion
    let pointer = Box::new(block).into_raw();
```

enzoc4: I don't think I can do that, the compiler tells me that `into_raw` is an associated function, not a method?

And according to the Rust documentation:
> Note: this is an associated function, which means that you have to call it as Box::into_raw(b) instead of b.into_raw(). This is so that there is no conflict with a method on the inner type.

PR #640, opened 2019-04-16
network-grpc: URI path-and-query part is required

mzabaluev:
Fix a crash when connecting.


NicolasDP:
Should we fill a bug report ? Sounds like this should be the default value. No ? Or at least API should not allow building it without the needed Parameters

mzabaluev:
> Should we fill a bug report ? Sounds like this should be the default value. No ?

I don't know, it's probably by design.

PR #642, opened 2019-04-23
[IOHKCRB-23] Block header functions

enzoc4:
---
name: [IOHKCRB-23] Block header functions
about: ^

---
**This PR is branched from https://github.com/input-output-hk/rust-cardano/pull/639, it shouldn't be merged before that one.**

# Description

## Sumary
Add functions to work with block headers

- `cardano_block_get_header`
- `cardano_raw_block_header_decode`
    (meant to decode the _/:network/tip_ endpoint of the [http-bridge](https://github.com/input-output-hk/cardano-http-bridge#get-networktip))
- `cardano_block_header_previous_hash`
- `cardano_block_header_compute_hash`
- `cardano_block_delete_hash`
- `cardano_block_header_delete`

## Motivation
Currently, there is no way to work with the result of the **/:network/tip** (because it is a `BlockHeader` and not a `Block`) endpoint on the http-bridge outside the Rust library.

There is also no way of traversing the blockchain, because even if you can get a block by its id (with the http bridge), there is no way of getting the id of the previous one.

The purpose of this functions is to do the following: 
- get tip from http bridge → decode it with C bindings → hash it → use that hash to get the block body from the http bridge
- get block from http bridge → decode it → get header → get hash of previous block → get previous block from http bridge


cardano-c/test/test.c
```diff
@@ -58,6 +58,8 @@ void valid_entropy_size_returns_success()
     {
         size_t size = valid_sizes[i];
         uint8_t *valid_wallet_entropy = malloc(size);
+        //This memset is to avoid valgrind to complain on use of unitialized memory
+        memset(valid_wallet_entropy, 0, size);
```
vincenthz: might have been to switch malloc to calloc instead of using memset

PR #645, opened 2019-04-25
Basic update proposal handling

edolstra:
Changes of note:

* Update proposals now need to be signed by a BFT leader.

* Adds a `UpdateVote` message type that must be signed by a BFT leader and registers a positive vote for a proposal. Proposal IDs are the message ID of the `UpdateProposal`.

* When proposals reach a majority, they become active at the start of the next epoch. (This should be changed.) If they don't get a majority after `Setting::proposal_expiration` epochs, they're expired.

* `Ledger::apply_block()` now checks the date and chain length of the incoming block.

* Fixes a critical bug in `Multiverse::get_from_storage()`.

### DIFF
--- a/chain-impl-mockchain/src/setting.rs
+++ b/chain-impl-mockchain/src/setting.rs
@@ -24,27 +24,38 @@ impl UpdateState {
         }
     }
 
-    pub fn apply_vote(mut self, vote: &SignedUpdateVote) -> Result<Self, Error> {
-        match vote.verify() {
-            Verification::Failed => Err(Error::BadVoteSignature(
+    pub fn apply_vote(
+        mut self,
+        vote: &SignedUpdateVote,
+        settings: &Settings,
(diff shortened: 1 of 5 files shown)


chain-impl-mockchain/src/setting.rs
```diff
@@ -159,6 +303,160 @@ impl Readable for UpdateProposal {
...
+        Box::new(std::iter::once(&self.proposer_id.0))
+    }
+}
+
+impl property::Serialize for UpdateProposalWithProposer {
+    type Error = std::io::Error;
+    fn serialize<W: std::io::Write>(&self, writer: W) -> Result<(), Self::Error> {
+        use chain_core::packer::*;
+        let mut codec = Codec::from(writer);
+        self.proposal.serialize(&mut codec)?;
```
vincenthz: minor point, but I don't think you don't need to create a codec and can pass the writer directly here

PR #648, opened 2019-04-27
add block0 in the verification data of the transaction witnesses

### LINKED ISSUE #644: sign transaction with the block0 hash

NicolasDP:
fix #644
relates to input-output-hk/jormungandr#292

### DIFF
--- a/chain-impl-mockchain/src/ledger.rs
+++ b/chain-impl-mockchain/src/ledger.rs
@@ -351,6 +351,7 @@ fn internal_apply_transaction(
             InputEnum::AccountInput(account_id, value) => {
                 ledger.accounts = input_account_verify(
                     ledger.accounts,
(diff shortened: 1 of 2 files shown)


chain-impl-mockchain/src/ledger.rs
```diff
@@ -456,7 +457,9 @@ fn input_utxo_verify(
                 ));
             };
 
-            let verified = signature.verify(&xpub, &transaction_id);
+            let data_to_verify =
+                WitnessUTxOData::new(&ledger.static_params.block0_initial_hash, &transaction_id);
```
mzabaluev: I'd rather rename all `UTxO` to `Utxo`: more in line with Rust naming conventions for acronyms and less keyboard twisting.

NicolasDP: good spot. Thanks

PR #652, opened 2019-04-29
Revert #645

edolstra:
It broke transaction handling in jormungandr because it uses apply_block to test whether transactions can be applied, but apply_block wants to update the chain_length / block date.


NicolasDP:
But it is fine since the new created state is throw away.

The transaction validation is done as follow:

1. take what is believed to be the tip of the blockchain;
2. apply the transaction;
3. throw the newly created state

PR #663, opened 2019-05-02
Replace trait Deserialize with trait Readable in network protocols

mzabaluev:
It seems `Readable` is what the cool kids use these days, and the impl of `Deserialize` for mockchain's `Header` was wrong anyway as it expected the length-prefixed format, contrary to what `Serialize` produces for the same type.

### DIFF
--- a/chain-core/src/mempack.rs
+++ b/chain-core/src/mempack.rs
@@ -165,10 +165,8 @@ impl<'a> ReadBuf<'a> {
     }
 }
 
-pub trait Readable {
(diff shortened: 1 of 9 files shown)


chain-core/src/mempack.rs
```diff
@@ -165,10 +165,8 @@ impl<'a> ReadBuf<'a> {
     }
 }
 
-pub trait Readable {
-    fn read<'a>(buf: &mut ReadBuf<'a>) -> Result<Self, ReadError>
-    where
-        Self: Sized;
```
NicolasDP: better, indeed

PR #671, opened 2019-05-06
Update tower-grpc dependency

mzabaluev:
tower-grpc-build now requires HTTP stack selection feature to generate support code for the stack.
Will fix input-output-hk/jormungandr#323.


sjmackenzie:
Great! Now just need for some kind soul to update `jormungandr's` `cardano-deps`. Thanks @mzabaluev.

PR #680, opened 2019-05-13
small changes in APIs linked to time

NicolasDP:
also add the missing link to configure the number of blocks per epoch from the block0 (the config parameter was already there).

### DIFF
--- a/chain-impl-mockchain/src/multiverse.rs
+++ b/chain-impl-mockchain/src/multiverse.rs
@@ -291,6 +291,7 @@ mod test {
 
     #[test]
     pub fn multiverse() {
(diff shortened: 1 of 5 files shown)


chain-impl-mockchain/src/multiverse.rs
```diff
@@ -291,14 +291,15 @@ mod test {
 
     #[test]
     pub fn multiverse() {
+        const NUM_BLOCK_PER_EPOCH: u32 = 1000;
```
mzabaluev: Bikeshedding: I'd name it `BLOCKS_PER_EPOCH`.

PR #682, opened 2019-05-13
network: Redesign block subscription for upload

mzabaluev:
Block subscription stream now can retrieve solicitations for blocks from the client, which can fulfill the solictitation with client streaming method `UploadBlocks`.

### DIFF
--- a/network-core/src/subscription.rs
+++ b/network-core/src/subscription.rs
@@ -1,0 +1,25 @@
+use chain_core::property::{Block, HasHeader};
+
+use std::fmt::{self, Debug};
+
+pub enum BlockEvent<B>
+where
+    B: Block + HasHeader,
+{
+    Announce(<B as HasHeader>::Header),
+    Solicit(Vec<<B as Block>::Id>),
+}
+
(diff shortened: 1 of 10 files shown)


network-core/src/subscription.rs
```diff
@@ -0,0 +1,25 @@
...
+impl<B> Debug for BlockEvent<B>
+where
+    B: Block + HasHeader,
+    <B as HasHeader>::Header: Debug,
+    <B as Block>::Id: Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            BlockEvent::Announce(header) => f.debug_tuple("Announce").field(header).finish(),
+            BlockEvent::Solicit(ids) => f.debug_tuple("Solicit").field(ids).finish(),
```
NicolasDP: nice, I never really looked into it before (the `debug` feature)

mzabaluev: It's what `#[derive(Debug)]` does automatically, but here it needs to be done manually because of how the type parameter is not part of the contents, but its associated types are.

PR #689, opened 2019-05-15
Add ability of decoding PublicKey from hex

CodeSandwich:
This is needed so https://github.com/input-output-hk/jormungandr/issues/328 can accept hex-encoded account IDs.

Also fixes small bug of hex decoder ignoring last valid hex character when input has uneven number of them

### DIFF
--- a/chain-crypto/src/key.rs
+++ b/chain-crypto/src/key.rs
@@ -3,6 +3,7 @@ use crate::hex;
 use rand::{CryptoRng, RngCore};
 use std::fmt;
 use std::hash::Hash;
+use std::str::FromStr;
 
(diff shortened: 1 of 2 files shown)


chain-crypto/src/key.rs
```diff
@@ -91,9 +109,30 @@ impl fmt::Display for PublicKeyError {
         }
     }
 }
+
+impl fmt::Display for PublicKeyFromStrError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            PublicKeyFromStrError::HexMalformed(_) => "Hex encoding malformed",
```
mzabaluev: The message strings should be lower case.

PR #691, opened 2019-05-16
network-grpc: #[must_use] on futures and streams

mzabaluev:
The compiler will warn if a returned future or stream value is dropped
without being used.


NicolasDP:
excellent

PR #693, opened 2019-05-17
Upstreaming OBFT stuff

vantuz-subhuman:
Upstreaming all the changes we have done in the fork. Includes:
1. OBFT processing (**does not include OBFT signature validation**)
2. Rollback handling
3. Special new index for epoch-lengths and loose blocks
4. Changes in epoch pack files to include epoch length and flags for whether it's an EBB epoch or not
5. Changes in storage to support `block-by-height` and `status` endpoints in the bridge ("status" returns information about currently known local and network tips)


cardano/src/block/verify_chain.rs
```diff
@@ -14,20 +14,34 @@ impl ChainState {
...
+                    let block_epoch = blk.header.consensus.slot_id.epoch;
+                    let local_epoch = self.last_date.map(|d| d.get_epochid()).unwrap_or(0);
+                    if block_epoch > local_epoch {
+                        // We are in OBFT and switched epochs without EBB
+                        // Remove last boundary block state
+
+                        // TODO: should cleanup EBB state here?
+                        //self.last_boundary_block = None;
+                        //self.last_boundary_block_epoch = None;
+                        //self.slot_leaders = vec![];
```
vantuz-subhuman: TODO: gotta decide whether we need to drop "Last EBB" state here after switch to OBFT. Cuz right now it just hangs there forever.

ℹ️ Gotta remember that "Last EBB" state is used to determine whether current epoch is an EBB or not, when storing epoch packs with meta-flags.

cardano/src/block/verify_chain.rs
```diff
@@ -65,17 +79,21 @@ impl ChainState {
...
-                    return Err(Error::BlockDateInFuture);
+                // New date should be this or next epoch, not further
+                if date.get_epochid() > last_date.get_epochid() {
+                    if date.get_epochid() > last_date.get_epochid() + 1 {
+                        return Err(Error::BlockDateInFuture);
+                    }
+                    // First next-epoch block supposed to be EBB,
+                    // unless we are in OBFT era
+                    if !date.is_boundary() {
+                        // TODO: validate we are in OBFT?
```
vantuz-subhuman: TODO: gotta decide whether we need to add additional checks for proper OBFT switch (e.g. block version, or some other weirdo parameters from blocks). Right now we just allow new epochs to have no EBB, and that's it.

cardano/src/block/verify_chain.rs
```diff
@@ -87,19 +105,33 @@ impl ChainState {
...
+                        return Err(Error::WrongSlotLeader);
+                    }
+                } else {
 
-                // Note: the block signature was already checked in
-                // verify_block, so here we only check the leader key
-                // against the genesis block.
-                if slot_leader != &address::StakeholderId::new(&blk.header.consensus.leader_key) {
-                    return Err(Error::WrongSlotLeader);
+                    // TODO: validate OBFT leader
```
vantuz-subhuman: No OBFT leader validation present. We don't have the documentation yet on how it's done.

storage/src/epoch.rs
```diff
@@ -1,30 +1,31 @@
...
+    index: indexfile::Index,
+    flags: &EpochFlags,
 ) {
     let dir = config.get_epoch_dir(epochid);
     fs::create_dir_all(dir).unwrap();
 
-    let pack_filepath = config.get_epoch_pack_filepath(epochid);
-    tmpfile::atomic_write_simple(&pack_filepath, hex::encode(packref).as_bytes()).unwrap();
+    epoch_write_pack(config, packref, &None, epochid, index.offsets, flags).unwrap();
+    // TODO: need to put new entry to storage, but storage is not present here =(
```
vantuz-subhuman: ⚠️ Idk what to do with this function, cuz by function definition we don't have all the context we need in here, but this function is never actually called anywhere, so we gotta either just delete it, or maybe add required context to function arguments, so if it's ever called - caller will be aware of it.

vantuz-subhuman:
Note to self:
- [ ] Check and fix if needed loose blocks deleting from file-system at rollback

vantuz-subhuman:
@disassembler

exe-common/src/config.rs
```diff
@@ -342,6 +342,28 @@ pub mod net {
             }
         }
 
+        pub fn obft_testnet() -> Self {
```
NicolasDP: We don't want to embed obft testnet in the library

vantuz-subhuman: Makes sense

vantuz-subhuman: Will remove then. Along with the weird hack for genesis.

vantuz-subhuman: It seems like we don't need obft-testnet anymore either, since staging seems to be working fine with OBFT. So yes. No reason to keep it.

vantuz-subhuman:
Notes to self:
- [ ] Remove obft-testnet (**along with the weird genesis hash hack**)
- [ ] Check and fix if needed loose blocks deleting from file-system at rollback

vantuz-subhuman:
Discovered that we, after all, need to fix loose blocks deleting at rollbacks, and also need to close some rollback scenarios that we made possible in the code, but which are impossible in Cardano. @rooooooooob is working on it atm. Will activate PR after this.

vincenthz:
@vantuz-subhuman there's no reason to explicitly delete loose blocks that are not used anymore, unless you really really want to delete them now. the storage design is that eventually this is taken care by packing/gc (think of git)

vantuz-subhuman:
> the storage design is that eventually this is taken care by packing/gc (think of git)

🤔 Interesting. I thought loose blocks only deleted when epoch becomes stable and gets packed, but then we go from tip back in history to link the epoch blocks, so blocks that were forked out would get lost.

Seems I might be missing something there then. Will check it out, thank you! But then we are still waiting for some small fixes in rollback logic, to not allow it to go ways it shouldn't.

vincenthz:
All the code to do that is not necessary in place (can't remember the details right now), but it's not a problem to have unreferenced blocks on disk, just like a commit in git that is not used anymore, it is eventually gc

vincenthz:
I started looking at the PR, but this likely suffer the same too-many-things-at-once problem of the first one.

Please cut the obvious improvements, refactoring and adjustment into multiples PRs, and this would help massively get this push through

PR #694, opened 2019-05-17
Ledger unit tests

dkijania:


dkijania:
Right now all tests cases are in ledger.rs. IMHO test code for ledger unit tests is so large that should be promoted to new module like below:

--ledger
-----ledger.rs
-----tests.rs
-----mod.rs

but this requires changes in production code.

chain-impl-mockchain/src/ledger.rs
```diff
@@ -865,77 +895,299 @@ pub mod test {
...
+            })
+            .finalize()
+            .with_utxo_witness(&block0_hash, &sk1)
+            .seal();
+
+        let dyn_params = ledger.get_ledger_parameters();
+        let r = ledger.apply_transaction(&signed_tx, &dyn_params);
+        assert!(r.is_ok());
+    }
+    struct TransactionBuilder {
```
vincenthz: This is great, but would it make sense to either unify with an existing functionality or move this to a general module ?

chain-impl-mockchain/src/ledger.rs
```diff
@@ -811,6 +812,34 @@ pub mod test {
...
+    ) -> (SpendingSecretKey, SpendingPublicKey, Address) {
+        let sk = SpendingSecretKey::generate(rng);
+        let pk = sk.to_public();
+        let user_address = Address(discrimination.clone(), Kind::Account(pk.clone()));
+        (sk, pk, user_address)
+    }
+
+    pub fn make_utxo_delegation_key<R: RngCore + CryptoRng>(
+        rng_single: &mut R,
+        rng_delegation: &mut R,
```
vincenthz: I don't think you need to create 2 rng here, you can reuse 1 multiple times

NicolasDP:
Regarding moving the tests to a different directory. there is a directory specific for this: `tests`. And it will consume the crate from a library point of view. Which is more or less what we want to do anyway.

vincenthz:
merging now, to not let this to rot and we can address the comment in separate future PRs, including moving this to another place

PR #697, opened 2019-05-20
Fix above_stake_threshold and add genesis leadership selection test

edolstra:
This code is primarily to experiment with the phi function, so a test is not really an ideal place for it. Might be better as an example?

### DIFF
--- a/chain-time/src/timeframe.rs
+++ b/chain-time/src/timeframe.rs
@@ -6,7 +6,7 @@ use std::time::{Duration, SystemTime};
 /// The slots are not comparable to others slots made on a
 /// different time frame
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
-pub struct Slot(pub u64);
+pub struct Slot(pub(crate) u64);
 
 impl From<Slot> for u64 {
     fn from(s: Slot) -> u64 {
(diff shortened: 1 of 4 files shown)


vincenthz:
Maybe best to see this as an executable in a separate crate ?
not sure what's the best way to access some of the function like the `create_fake_ledger`

edolstra:
Apparently it's possible to disable a test by default by marking it with `#[ignore]`. That sounds like a good option here.

chain-time/src/timeframe.rs
```diff
@@ -6,7 +6,7 @@ use std::time::{Duration, SystemTime};
 /// The slots are not comparable to others slots made on a
 /// different time frame
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
-pub struct Slot(pub(crate) u64);
+pub struct Slot(pub u64);
```
NicolasDP: does it need to be public? I think this value, the only way we should get it, should be through the `TimeFrame` object. It's a bit annoying but this is the only way to guarantee this object is not constructed with the wrong value, or with value that does not make sense in a given timeframe.

edolstra: I've fixed the test to use `ledger.settings.era` so it doesn't need to construct `Slot` anymore.

PR #699, opened 2019-05-21
Ledger unit test - fix test organization

dkijania:


chain-impl-mockchain/src/block/mod.rs
```diff
@@ -8,9 +8,9 @@ use std::slice;
 
 mod builder;
 //mod cstruct;
-mod header;
+pub mod header;
 mod headerraw;
-mod version;
+pub mod version;
```
NicolasDP: this should not be necessary because we expose all the interfaces in the `pub use self::version::*` below

chain-impl-mockchain/src/transaction/mod.rs
```diff
@@ -1,7 +1,7 @@
-mod transaction;
-mod transfer;
-mod utxo;
-mod witness;
+pub mod transaction;
+pub mod transfer;
+pub mod utxo;
+pub mod witness;
```
NicolasDP: We should not need to make this interface public, allthe APIs are re-exported below with `pub use self::transaction::*;` and following lines.

dkijania: Done

chain-impl-mockchain/src/block/mod.rs
dkijania: Done

chain-impl-mockchain/tests/ledger/ledger_tests.rs
```diff
@@ -0,0 +1,232 @@
...
+use chain_addr::Discrimination;
+
+use chain_impl_mockchain::account::Identifier;
+use chain_impl_mockchain::account::SpendingCounter;
+
+use chain_impl_mockchain::ledger::Error::NotEnoughSignatures;
+use chain_impl_mockchain::transaction::*;
+use chain_impl_mockchain::value::*;
+
+macro_rules! assert_err {
```
mzabaluev: Could just use `.expect_err()` and examine the returned error with `assert_eq!`.

PR #703, opened 2019-05-23
New exposed accessor functions

vantuz-subhuman:
Added some accessors here and there to read values hidden at the moment. And added couple sugar accessors.


vantuz-subhuman:
@vincenthz @NicolasDP @disassembler 

https://github.com/input-output-hk/rust-cardano/pull/693#issuecomment-494574901

PR #708, opened 2019-05-29
key evolution cleanup

vincenthz:


chain-crypto/src/algorithms/sumed25519/mod.rs
```diff
@@ -87,16 +88,21 @@ impl SigningAlgorithm for SumEd25519_12 {
     }
 }
 
+impl KeyEvolvingAlgorithm for SumEd25519_12 {
+    fn get_period(sec: &Self::Secret) -> usize {
```
mzabaluev: What is the period measured in? For values not actually tied to memory size, it's preferable to use `u64`, lest our innumerable fan base with Raspberry Pis gets hit with integer overflows.

chain-crypto/src/algorithms/sumed25519/mod.rs
```diff
@@ -87,16 +88,21 @@ impl SigningAlgorithm for SumEd25519_12 {
...
-    fn update(key: &mut Self::Secret) -> bool {
-        sum::update(key).is_ok()
-    }
-    fn sign_update(key: &mut Self::Secret, msg: &[u8]) -> Self::Signature {
-        let sig = sum::sign(key, msg);
-        let _ = sum::update(key);
-        sig
-    }
-}
+}
```
mzabaluev: rustfmt this into having properly terminated lines, please.

chain-crypto/src/algorithms/sumed25519/mod.rs
vincenthz: I moved the interface to use `u32` for now because a proper fix is hairy, but it's neither the right fix nor that it matter much at this stage.

mzabaluev: OK. The general rule of thumb for integer types that are not memory object sized is to use `i64`/`u64` unless the value is known to be small enough to fit in smaller sized types, including any possible intermediate expression values.

PR #719, opened 2019-06-05
Add MemoryBlockStore and SQLiteBlockStore tests

edolstra:


chain-storage-sqlite/src/lib.rs
```diff
@@ -183,17 +183,24 @@ where
...
-            .unwrap();
-        Ok(())
+            ]) {
+            Ok(_) => Ok(()),
+            Err(rusqlite::Error::SqliteFailure(err, _))
+                if err.code == rusqlite::ErrorCode::ConstraintViolation =>
+            {
+                Err(Error::BlockNotFound)
+            }
+            Err(err) => panic!(err),
```
mzabaluev: Don't want to add a catch-all error variant at this point?

chain-storage/Cargo.toml
```diff
@@ -11,7 +11,9 @@ edition = "2018"
 
 [dependencies]
 chain-core = { path = "../chain-core" }
+rand = "0.6"
 
 [dev-dependencies]
+#rand = "0.6"
```
mzabaluev: Just remove this line.

chain-storage/src/store.rs
```diff
@@ -298,3 +299,270 @@ fn compute_fast_link(depth: u64) -> u64 {
         0
     }
 }
+
+//#[cfg(test)]
```
mzabaluev: Why is this now enabled in non-test configurations?
If substantial additional modules are needed to support the tests, it's better to put them in subdirectories under `tests` in the crate directory, and organize the test code into standalone tests.

chain-storage/Cargo.toml
mzabaluev: Actually, if `rand` is only used in test code, the reorganization described in my previous comment would allow it to be kept as a dev-dependency.

chain-storage/src/store.rs
```diff
@@ -300,7 +300,7 @@ fn compute_fast_link(depth: u64) -> u64 {
     }
 }
 
-//#[cfg(test)]
```
mzabaluev: Elsewhere we put the public test API under mod `testing`, distinct from mod `tests` which is conventional for the unit tests. let's follow the same convention.

chain-storage/src/error.rs
```diff
@@ -4,13 +4,15 @@ use std::{error, fmt};
 pub enum Error {
     BlockNotFound, // FIXME: add BlockId
     CannotIterate,
+    BackendError,
```
mzabaluev: Good. It could also pack a `Box<dyn Error + Send + Sync>` with the underlying error that would be chained in the `Error` impl, but this can be done later.

edolstra: Done, thanks!

chain-storage/src/memory.rs
```diff
@@ -71,3 +75,27 @@ where
         self as &BlockStore<Block = Self::Block>
     }
 }
+
+#[cfg(test)]
+pub mod test {
```
mzabaluev: nits:
- does not need to be public;
- the conventional name is `tests`.

edolstra: Fixed. BTW we have 42 modules named `test` so we're not following the convention very well :-)

chain-storage/src/error.rs
```diff
@@ -1,16 +1,18 @@
 use std::{error, fmt};
 
-#[derive(Debug, Clone, Eq, PartialEq)]
+#[derive(Debug)]
 pub enum Error {
     BlockNotFound, // FIXME: add BlockId
     CannotIterate,
+    BackendError(Box<dyn std::error::Error + Send + Sync>),
```
vincenthz: is it worth it to keep the error in the native format (and lose the derives), instead of flatting it into something simple ? The fact that it's handed this way just in the next chunk makes me think that just having a String of what this internal error is about, will be useful enough for debugging

mzabaluev: It's not typical for errors to implement `PartialEq`, precisely because of the error information baggage. If we want to be really thorough, we sould have an `ErrorKind` enum as a property, and maybe switch to `snafu` for all-bells-and-whistles error utilities.

NicolasDP: I agree with @mzabaluev , this is error chaining and there are lot of package/standard way to do this kind of thing (like [error-chain](https://crates.io/crates/error-chain))

PR #724, opened 2019-06-11
Adding a utility that'll allow the conversion of one address

sjmackenzie:
to another address where the human readable prefix changes.


sjmackenzie:
lifted up into https://github.com/input-output-hk/jormungandr/pull/453/commits/6266dc57165d9fb793c82fbd834c10c0a2800466, hence no need for a separate function, as it's so small.

Issue #728, opened 2019-06-11
Testing Lvl1: checking that the transaction builder build valid transaction


NicolasDP:
We need to validate that the transaction builder is actually building valid transactions.

utilising `quickcheck`, write property tests to validate the building of the transaction is correct.

For example:

* for an arbitrary set of inputs (len <= 255);
* for an arbitrary set of outputs (len <= 256);
* for an arbitrary set of fee parameters

build a transaction

* the result should give a valid, balanced transaction;

_it is important to note that the output values is impacted by the total input available and the fee parameters_.

relates to epic ticket: #727

PR #733, opened 2019-06-13
Add property test to TransactionBuilder

CodeSandwich:
Fixes https://github.com/input-output-hk/rust-cardano/issues/728

### DIFF
--- a/chain-impl-mockchain/Cargo.toml
+++ b/chain-impl-mockchain/Cargo.toml
@@ -29,6 +29,7 @@ custom_error = "1.6"
 
 [dev-dependencies]
 quickcheck = "0.8"
+quickcheck_macros = "0.8.0"
 chain-core = { path = "../chain-core", features=["property-test-api"]}
 chain-crypto = { path = "../chain-crypto", features=["property-test-api"]}
 chain-addr = { path = "../chain-addr", features=["property-test-api"]}
--- a/chain-impl-mockchain/src/transaction/transfer.rs
+++ b/chain-impl-mockchain/src/transaction/transfer.rs
@@ -6,7 +6,7 @@ use chain_core::mempack::{ReadBuf, ReadError, Readable};
 use chain_core::property;
 use chain_crypto::PublicKey;
 
-const INPUT_PTR_SIZE: usize = 32;
+pub const INPUT_PTR_SIZE: usize = 32;
 
 /// Generalized input which have a specific input value, and
 /// either contains an account reference or a TransactionId+index
--- a/chain-impl-mockchain/src/txbuilder.rs
+++ b/chain-impl-mockchain/src/txbuilder.rs
@@ -43,7 +43,7 @@ pub enum OutputPolicy {
     Forget,
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 /// Transaction builder is an object to construct
 /// a transaction with iterative steps (inputs, outputs)
 pub struct TransactionBuilder<Address, Extra> {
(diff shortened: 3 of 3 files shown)


chain-impl-mockchain/src/transaction/transfer.rs
```diff
@@ -6,7 +6,7 @@ use chain_core::mempack::{ReadBuf, ReadError, Readable};
 use chain_core::property;
 use chain_crypto::PublicKey;
 
-const INPUT_PTR_SIZE: usize = 32;
+pub const INPUT_PTR_SIZE: usize = 32;
```
NicolasDP: why do you need this public?

CodeSandwich: Because we need to generate an input pointer. This probably should be public in the first place, because it's a part of `Input` and `AccountIdentifier` public APIs

chain-impl-mockchain/src/txbuilder.rs
```diff
@@ -289,3 +289,136 @@ impl TransactionFinalizer {
...
+    quickcheck! {
+        fn tx_builder_creates_balanced_tx(arbitrary_builder: ArbitraryTransactionBuilder) -> () {
+            let builder = arbitrary_builder.0;
+            let policy = OutputPolicy::Forget;
+            let fee = fee_for_tx_builder(&builder);
+
+            let (balance, _transaction) =  builder.finalize(fee, policy).unwrap();
+
+            match balance {
+                Balance::Positive(_) => panic!("Positive balance"),
```
NicolasDP: don't use panic, quickcheck as a type `TestResult` in which you can provide information

chain-impl-mockchain/src/txbuilder.rs
```diff
@@ -289,3 +289,136 @@ impl TransactionFinalizer {
...
+    use crate::transaction::Input;
+    use crate::transaction::NoExtra;
+    use crate::transaction::INPUT_PTR_SIZE;
+    use chain_addr::Address;
+    use quickcheck::Arbitrary;
+    use quickcheck::Gen;
+    use std::ops::RangeBounds;
+
+    quickcheck! {
+        fn tx_builder_creates_balanced_tx(arbitrary_builder: ArbitraryTransactionBuilder) -> Result<(), &'static str> {
```
NicolasDP: I was speaking of this: https://docs.rs/quickcheck/0.8.5/quickcheck/struct.TestResult.html

CodeSandwich:
@NicolasDP I implemented `the result should give a valid, balanced transaction;` part a bit too literally ;) It's fixed now, transactions are built randomly and test verifies if builder accepts correct combinations.

chain-impl-mockchain/src/txbuilder.rs
```diff
@@ -289,3 +289,153 @@ impl TransactionFinalizer {
         }
     }
 }
+
+#[cfg(test)]
```
dkijania: consider making this test as public interface test (by moving them to `rust-cardano/tests`). Thanks to that we can split its logic across separate files.

chain-impl-mockchain/src/txbuilder.rs
```diff
@@ -289,3 +289,153 @@ impl TransactionFinalizer {
...
+    use crate::fee::LinearFee;
+    use crate::transaction::{Input, INPUT_PTR_SIZE};
+    use chain_addr::Address;
+    use quickcheck::{Arbitrary, Gen, TestResult};
+    use quickcheck_macros::quickcheck;
+    use rand::distributions::uniform::{SampleUniform, Uniform};
+    use rand::Rng;
+
+    #[quickcheck]
+    fn tx_builder_never_creates_unbalanced_tx(
```
dkijania: IMHO this method is too long and has exposed too much of it's logic

chain-impl-mockchain/src/txbuilder.rs
```diff
@@ -289,3 +289,153 @@ impl TransactionFinalizer {
...
+        let input_sum = Value::sum(inputs.0.iter().map(|input| input.value)).unwrap();
+        let output_sum = Value::sum(outputs.0.iter().map(|output| output.1)).unwrap();
+        let fee_value = fee.calculate(&builder.tx).unwrap_or(Value::zero());
+        let is_accepted = (input_sum - output_sum)
+            .and_then(|balance| balance - fee_value)
+            .is_ok();
+
+        let result = builder.finalize(fee, OutputPolicy::Forget);
+
+        if is_accepted {
```
dkijania: this (from 327-345) can be extracted to a method

chain-impl-mockchain/src/txbuilder.rs
CodeSandwich: Do you mean the whole test module or just the helper functions?

chain-impl-mockchain/Cargo.toml
```diff
@@ -29,6 +29,7 @@ custom_error = "1.6"
 
 [dev-dependencies]
 quickcheck = "0.8"
+quickcheck_macros = "0.8.0"
```
NicolasDP: why do you need this one now?

CodeSandwich: We don't need it, but it makes our lives a bit easier. Macro by example prevents rustfmt from formatting test bodies unlike proc macro. This dependency is low cost too, it itself doesn't have any dependencies not covered by serde_derive anyway.

chain-impl-mockchain/src/txbuilder.rs
dkijania: I meant the whole test module,

CodeSandwich: I'm not sure about that. This test touches only TransactionBuilder and nothing else, it's more of a unit test than integration test. If we keep it close to TransactionBuilder, it will be much easier to find.

dkijania: sure, this is good reason

PR #735, opened 2019-06-13
Ledger serialization/deserialization

edolstra:
This adds serialization/deserialization of `Ledger` using serde, which can be used by jormungandr to save/restore the chain state periodically.

It also adds `PartialEq` and `Eq` to `Ledger` and its component types. This is only used by the tests to determine if `Ledger` is restored correctly.

### DIFF
--- a/imhamt/src/hamt.rs
+++ b/imhamt/src/hamt.rs
@@ -242,7 +242,9 @@ impl<H: Default + Hasher, K: Eq + Hash, V: PartialEq> PartialEq for Hamt<H, K, V
         }
         for (k, v) in self.iter() {
             if let Some(v2) = other.lookup(k) {
-                if v != v2 { return false; }
+                if v != v2 {
+                    return false;
+                }
             } else {
                 return false;
             }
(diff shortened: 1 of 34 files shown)


imhamt/src/hamt.rs
```diff
@@ -202,3 +202,53 @@ impl<H: Default + Hasher, K: Eq + Hash, V> FromIterator<(K, V)> for Hamt<H, K, V
...
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let map = std::collections::HashMap::<K, V>::deserialize(deserializer)?;
+        Ok(Self::from_iter(map.into_iter()))
+    }
+}
+
+impl<H: Default + Hasher, K: Eq + Hash, V: PartialEq> PartialEq for Hamt<H, K, V> {
+    fn eq(&self, other: &Self) -> bool {
```
edolstra: @vincenthz If it's guaranteed that two equal HAMTs have identical in-memory representation, we could also just use ` derive Eq` for `Hamt`.

vincenthz: it's supposed to be guaranteed yes (although not really tested), although not adding the Eq was on purpose since that offers a good way to do an operation that probably shouldn't be done

vincenthz:
@edolstra we purposely stayed away from Serde and its automatic (and opaque) derivations, and I think this is important to stay away from a code that make automatic/opaque choice for us, since it does usually have a higher cost down the line

Also imhamt has an iterator, which I think should be enough to dump everything. probably a good idea to provide a from_iterator somewhere also.

PR #737, opened 2019-06-14
[Test] Ledger test for outputs exceeding 254

dkijania:


edolstra:
There is a merge conflict at the moment.

chain-impl-mockchain/src/ledger.rs
```diff
@@ -582,9 +585,17 @@ fn internal_apply_transaction(
     witnesses: &[Witness],
     fee: Value,
 ) -> Result<Ledger, Error> {
-    assert!(inputs.len() < 255);
-    assert!(outputs.len() < 255);
-    assert!(witnesses.len() < 255);
+    if inputs.len() >= 255 {
```
NicolasDP: the doc says 256 inputs: https://github.com/input-output-hk/rust-cardano/blob/master/chain-impl-mockchain/doc/format.md#common-structure

vincenthz: I think `< 255` is indeed 1 too short, pretty much a copy paste error

chain-impl-mockchain/src/ledger.rs
```diff
@@ -581,9 +584,17 @@ fn internal_apply_transaction(
     witnesses: &[Witness],
     fee: Value,
 ) -> Result<Ledger, Error> {
-    assert!(inputs.len() < 255);
-    assert!(outputs.len() < 255);
-    assert!(witnesses.len() < 255);
+    if inputs.len() > 256 {
```
mzabaluev: Please define these magic numbers as constants somewhere visible at the beginning of the module.

chain-impl-mockchain/src/ledger.rs
```diff
@@ -97,6 +97,9 @@ custom_error! {
         OldUtxoInvalidPublicKey { utxo: UtxoPointer, output: OutputOldAddress, witness: Witness } = "Old Transaction with invalid public key",
         AccountInvalidSignature { account: account::Identifier, witness: Witness } = "Account with invalid signature",
         MultisigInvalidSignature { multisig: multisig::Identifier, witness: Witness } = "Multisig with invalid signature",
+        TransactionHasTooManyInputs = "Transaction has more than 255 inputs",
+        TransactionHasTooManyOutputs = "Transaction has more than 255 outputs",
+        TransactionHasTooManyWitnesses = "Transaction has more than 255 witnesses",
```
mzabaluev: These messages should also use the constant definitions, or omit the numbers entirely.

chain-impl-mockchain/src/transaction/transfer.rs
```diff
@@ -189,6 +198,12 @@ pub struct Output<Address> {
     pub value: Value,
 }
 
+impl<Address: Readable> Output<Address> {
+    pub fn from(address: Address, value: Value) -> Self {
```
mzabaluev: This name is confusingly similar to `From::from`. Please name it either `new`, if it's the primary way to construct `Output`, or something descriptive like `from_address_and_value` if it's one of many constructors.

chain-impl-mockchain/tests/common/tx_builder.rs
```diff
@@ -25,25 +26,22 @@ impl TransactionBuilder {
...
-        &'a mut self,
-        outputs: Vec<Output<Address>>,
-    ) -> &'a mut TransactionBuilder {
+    pub fn with_outputs<'a>(&'a mut self, outputs: Vec<Output<Address>>) -> &'a mut Self {
         self.outputs.extend(outputs.iter().cloned());
         self
     }
 
-    pub fn finalize<'a>(&'a mut self) -> &'a mut TransactionBuilder {
+    pub fn finalize<'a>(&'a mut self) -> &'a mut Self {
```
mzabaluev: This seems like it leaves the builder open for further modification, while adding the result members alongside inputs and outputs that may fall out of sync later.

I would change the method to return `Transaction` by value and remove the internal members `transaction` and `transaction_id`.

chain-impl-mockchain/tests/ledger/ledger_tests.rs
```diff
@@ -35,198 +31,162 @@ macro_rules! assert_err {
...
-    let mut delegation_rng = rand::thread_rng();
-    let (sk1, _pk1, user1_address) =
-        accounts::make_utxo_delegation_key(&mut rng, &mut delegation_rng, &discrimination);
-    let (_sk2, _pk2, user2_address) = accounts::make_account_key(&mut rng, &discrimination);
+    let signed_tx = TransactionBuilder::new()
+        .with_input(Input::from_utxo(utxos[0]))
+        .with_output(Output::from(receiver.address.clone(), Value(42000)))
+        .finalize()
+        .with_witness(&block0_hash, &faucet)
+        .seal();
```
mzabaluev: Now there's both a (non-final) `.finalize()` and `.seal()`, what gives? :smiley:

chain-impl-mockchain/tests/common/tx_builder.rs
mzabaluev: I see now why; the final method is not actually this one, but `seal`.

To make it less confusing, it may be better to have a two-stage linear-typish builder design, where this method could be changed to

```rust
    pub fn transaction(&mut self) -> TransactionAuthenticator {
        let transaction = Transaction {
            inputs: self.inputs.clone(),
            outputs: self.outputs.clone(),
            extra: NoExtra,
        };
        TransactionAuthenticator {
            transaction,
            witnesses: Vec::new(),
            // ...
        }
    }
```

and then move the witness stuff and the `seal` method, renamed `finalize`, to `AuthenticatedTransactionBuilder`.

This may not be as nicely chainable as `&mut self` methods on the single instance, but it could be made chainable all the way through by all modifier methods changed to take `self` by value and return a modified moved value in `Self`. In test code, we should care more about convenience than performance.

mzabaluev:
> Google recovered

s/Google/Microsoft who owns us on GitHub/

chain-impl-mockchain/src/ledger.rs
```diff
@@ -581,9 +589,26 @@ fn internal_apply_transaction(
...
+            current: outputs.len(),
+        });
+    }
+
+    if witnesses.len() > MAX_TRANSACTION_WITNESSES_COUNT {
+        return Err(Error::TransactionHasTooManyWitnesses {
+            expected: MAX_TRANSACTION_WITNESSES_COUNT,
+            current: witnesses.len(),
+        });
+    }
```
mzabaluev: Solid. I'd name the `current` members `actual`.

chain-impl-mockchain/src/transaction/transfer.rs
```diff
@@ -92,6 +94,13 @@ impl Input {
         }
     }
 
+    pub fn from_account_pk(public_key: SpendingPublicKey, value: Value) -> Self {
```
mzabaluev: The abbreviation is a bit short unless the same is used everywhere.
`from_account_public_key`?

chain-impl-mockchain/src/transaction/transfer.rs
```diff
@@ -189,6 +198,12 @@ pub struct Output<Address> {
     pub value: Value,
 }
 
+impl<Address: Readable> Output<Address> {
+    pub fn from_address_and_value(address: Address, value: Value) -> Self {
```
mzabaluev: Actually, looking at the other constructors, `_and_value` seems out of place here, as they all have the second `Value` parameter. So it can be named `from_address` for the perfect feng shui.

chain-impl-mockchain/tests/ledger/ledger_tests.rs
```diff
@@ -35,198 +31,200 @@ macro_rules! assert_err {
...
+
+    let signed_tx = TransactionBuilder::new()
+        .with_input(Input::from_utxo(utxos[0]))
+        .with_output(Output::from_address_and_value(
+            receiver.address.clone(),
+            Value(42000),
+        ))
+        .finalize()
+        .with_witness(&block0_hash, &faucet)
+        .seal();
```
mzabaluev: This would look more logical if the `finalize` method was renamed to reflect the intermediate building phase, like:

```rust
    let signed_tx = TransactionBuilder::new()
        .with_input(Input::from_utxo(utxos[0]))
        .with_output(Output::from_address_and_value(
            receiver.address.clone(),
            Value(42000),
        ))
        .authenticate()
        .with_witness(&block0_hash, &faucet)
        .seal();
```

mzabaluev: (Also, I was wrong and the intermediate returned object can be actually mutably borrowed)

PR #738, opened 2019-06-14
allow transaction without inputs or outputs

NicolasDP:
the fee/balance will do the work

FYI @dkijania and @CodeSandwich


CodeSandwich:
Is transaction with no inputs, no outputs and linear fee without constant value valid? Will it start being accepted?

What about shredder transaction with inputs but no output, is it ever acceptable?

NicolasDP:
multiple cases:

1. no inputs, no outputs and no certificates are covered by the constant fee setting. They will be rejected;
2. everything else requires an exact balance. It needs to include fees and the necessary inputs/outputs

CodeSandwich:
Is the exact balance requirement already enforced? We still haven't solved the problem of positive balance with value lower than fee for addition of another output.

vincenthz:
@CodeSandwich it's enforced at the moment, although yes there's some case that are quite complicated.. which I personally think is endemic to the problem of using utxo instead of accounts

PR #742, opened 2019-06-19
chain-storage: Get rid of excessive boxing and dynamism

mzabaluev:
Remove the boxed closures and dynamic trait objects.
The BlockStore trait loses object safety, but it should never
have required it.

`get_path_to_nth_accessor` is removed from the trait methods and turned into a free helper fn `for_path_to_nth_accessor`.

### DIFF
--- a/chain-storage-sqlite/examples/import.rs
+++ b/chain-storage-sqlite/examples/import.rs
@@ -4,7 +4,7 @@ extern crate chain_storage;
 
 use cardano_storage::StorageConfig;
 use chain_core::property::{Block, BlockId};
-use chain_storage::store::BlockStore;
+use chain_storage::store::{self, BlockStore};
 use std::env;
 use std::str::FromStr;
 use std::time::Instant;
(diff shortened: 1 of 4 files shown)


edolstra:
Does this still work in jormungandr? I seem to remember it was needed there (see `NodeStorage` in `jormungandr/src/start_up/mod.rs`).

mzabaluev:
I need a rust-cardano port to recent master to try it out. I'm going to use an enum. Or perhaps it would be more elegant to restore object safety with the dynamic reference in the iterator.

edolstra:
Hm, how is that an improvement though? What's the problem with the dynamism?

mzabaluev:
Many things were boxed when they don't have to. I have now provided impls for both the static and boxed case.

edolstra:
BTW shouldn't this PR be moved to https://github.com/input-output-hk/chain-libs?

mzabaluev:
Moved to input-output-hk/chain-libs#1 :1st_place_medal:

PR #745, opened 2019-07-30
Update genesis data to new testnet

KtorZ:
Testnet was reset recently. This update the genesis file and its hash.

Cheers :gift_heart:


NicolasDP:
You need to fix the test for the genesis data too, I believe the content has changed:

https://github.com/input-output-hk/rust-cardano/blob/3a7ad93d637af280d5cffa7fd89661a71d45c157/exe-common/src/genesisdata/parse.rs#L154-L175

PR #746, opened 2019-07-30
Throw error on odd length

SebastienGllmt:
Currently no error is thrown if an extra 4 bits are added to a string.

Example: https://repl.it/repls/ObviousDemandingActivecell


NicolasDP (approved):

NicolasDP (approved):
LGTM

